PR #11, opened 2014-07-01
Setting response headers before calling WriteHeader.

pinscript:
As documented in the ResponseWriter type:

```
Header returns the header map that will be sent by WriteHeader.
Changing the header after a call to WriteHeader (or Write) has
no effect.
```


onur-verdo:
Webhook test comment on PR

PR #16, opened 2014-07-02
Wrap http.ResponseWriter to greatly increase middleware/logging accesibilty

zmarcantel:
The returning status is hidden in http.ResponseWriter, but it's easy to wrap.

This follows Martini and other packages practice and allows giving access to many variables.
I have only included .Status() as an example, but Martini has body length and a few others.

There is also an example of an extended logger working as a middleware using the .Status() function to color output based on status code.
This examples could easily extend to other middleware that "fork" handling based on a specific status such as:

1. monitoring
2. logging
3. intrusion detection
4. internal health checks
4. etc


manucorporat:
@zmarcantel this was implemented a long time ago, but thanks for your work. I am sorry for the inconveniences I am going to add your username to the AUTHORS file.

PR #36, opened 2014-07-03
instanciate templates at engine creation

zaibon:
this allow to use Delims() on HTMLTemplates before calling LoadHTMLTemplates()
so we can change delimmiters in templates.


deprecated.go
```diff
@@ -1,8 +1,9 @@
 package gin
 
 import (
-	"github.com/gin-gonic/gin/binding"
 	"net/http"
+
+	"github.com/Zaibon/gin/binding"
```
gnarlyfactory: just curious -- why was this moved?

manucorporat: GoImports in Sublimetext automatically separes the external packages, it looks like a good practice in Go. https://packagecontrol.io/packages/GoImports

Also, check out the sample code: https://golang.org/doc/code.html#Library

gnarlyfactory: make sense, and does look neater.  was just curious as to why this change was coming together with the actual flow change

manucorporat:
@zaibon Sorry for the delay. I am not going to merge this since HTMLTemplate does not longer exist.

If you want custom template settings just do this:
```go
router := gin.Default()
router.SetHTMLTemplate(yourTemplate())
// [...]

func yourTemplate() *template.Template {
    templ := template.New("").
    templ.Delims("[[", "]]")
    // ... more config
    return templ
}

func yourHandler(c *gin.Context) {
    c.HTML(200, "template", data) // < this will render your custom template
}
```

PR #179, opened 2014-12-15
Adding helper functions to router group for LINK and UNLINK.

jamie-stackhouse:
Minus the small reformat that Goimports automatically made with the structure of imports, this just adds a few helper methods to the Router Group that would help us as we are using them in our project.

If you want, I can remove the import reordering.

Thanks for Gin!

### DIFF
--- a/routergroup.go
+++ b/routergroup.go
@@ -5,9 +5,10 @@
 package gin
 
 import (
-	"github.com/julienschmidt/httprouter"
 	"net/http"
 	"path"
+
+	"github.com/julienschmidt/httprouter"
 )
 
 // Used internally to configure router, a RouterGroup is associated with a prefix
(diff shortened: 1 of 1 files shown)


javierprovecho:
@jamie-stackhouse sorry for delay

I applied gofmt and merged. Thank you.

manucorporat:
@itsjamie how do you use LINK and UNLINK?

itsjamie:
Xhr2 supports custom methods types.

manucorporat:
@itsjamie what does it mean?

for v1. I may drop the shortcut for `LINK` and `UNLINK`, and you will have to use router.Handle() directly:
```go
router := gin.New()
router.Handle("LINK", "/path", handler)
```

itsjamie:
It is listed as an extension to HTTP 1.0.


It is used in a REST API along with the Link header to describe a relationship between two resources.

manucorporat:
Well, I have decided to provide shortcuts only for standard HTTP methods. In addition, HTTP 1.0? does it still exist?

We have to be consistent, there are too many HTTP extensions, and we do not want add shortcuts for each one. For example the `webdav` standard defines verbs like: MOVE, LOCK, UNLOCK.
http://www.webdav.org/specs/rfc2518.html

If anyone wants to use weird methods, such as LINK, UNLINK, LOCK, UNLOCK, PING... they should use:
```go
router.Handle("LINK", "/path", handler)
```

how does it sound?

itsjamie:
Sure thing.

PR #182, opened 2014-12-21
Fixed issue allowing to spoof ClientIP()

nl5887:
The X-Forwarded-For can be used to spoof the real client ip. The middleware
introduced in this patch (which should only be used when having servers
in front of this servers) will filter all defined proxies (or local ip
addresses by default) and replace the RemoteAddr with the real client
ip.

### DIFF
--- a/context.go
+++ b/context.go
@@ -12,7 +12,9 @@ import (
 	"github.com/gin-gonic/gin/render"
 	"github.com/julienschmidt/httprouter"
 	"log"
+	"net"
 	"net/http"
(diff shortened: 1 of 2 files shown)


manucorporat:
This should be a core feature of Gin, not a middleware. Also, Gin should never modify `http.Request`, we work at different layers of abstraction.

https://github.com/golang/go/blob/d5630142fdba6280c47458cf30165d3d62d44fc3/src/net/http/request.go#L203-L210

PR #200, opened 2015-01-30
Specify utf-8 as content type charset on renderers

joiggama:
Content type charset is not specified for renderers, currently there's no way to configure it. I believe this might be a sane default.


javierprovecho:
I also agree with this. Thank you @joiggama

PR #212, opened 2015-02-08
Fixing typo in Form parameters example

rayrod2030:
Should be capital G in Get.


javierprovecho:
Thank you @rayrod2030 !

PR #213, opened 2015-02-08
Update context.go

techjanitor:
Add localhost to reservedLocalIps, for situations where it's sitting behind something like nginx on the same host.


javierprovecho:
Thank you @techjanitor ! I agree with you.

PR #217, opened 2015-02-09
Added support for unsigned integers in binding parameters of form posts....

ethankan:
... Also changed parsing of integer fields to take into account the size of the fields.


javierprovecho:
@ethankan thank you!

nmfzone:
Hey, is it possible to extends this functionality? For example for handling nullable.String from [nullable](https://github.com/Thor-x86/nullable)?

PR #230, opened 2015-03-04
colorful logger on windows

mattn:
Before

After


rogierlommers:
Cool!

javierprovecho:
@mattn thank you!

PR #231, opened 2015-03-04
Add customizable Realm for Basic authentication

frankbille:
Depending on the use case, it might be useful to be able to have different realms for different route groups.


matejkramny:
+1

javierprovecho:
Great! Thank you @frankbille!

PR #232, opened 2015-03-05
Using net/http constants instead of numeric values

tebeka:


javierprovecho:
Thank you @tebeka !

I agree with that. Some people prefer numbers but constants are readable too, and also specify they mean a HTTP status, instead of a random number.

tebeka:
Thanks!

Issue #238, opened 2015-03-12
Headers are always written => gives a warning on hijacked connections


quetz:
What subj says :) I have websocket handler that manages connection by itself and can't find a way to stop gin from writing headers. It happens in routergroup.go around line 60. context.Writer.WriterHeaderNow()

Is there any workaround, or maybe we need a flag like 'hijacked' in a context struct?

The warning in question is: "http: response.WriteHeader on hijacked connection"

manucorporat:
I think this is always fixed here: https://github.com/gin-gonic/gin/commit/251b73fc70c4f75487ec827be7b0c3a3e877e11a

can you confirm?

quetz:
Yeah, that fixes it. Thank you very much.

mephux:
@manucorporat I think something may have been added recently to cause this again.. not a header write but a normal response.Write - anyone else having a similar issue? /cc @quetz

quetz:
@mephux, not using Gin nowadays. Moved over to https://github.com/carbocation/interpose and http://www.gorillatoolkit.org/pkg/mux

mephux:
@quetz ah, ok.

For anyone coming to this threat I found the problem - was an issue with the gzip middleware. (https://github.com/gin-gonic/contrib/pull/71)

PR #239, opened 2015-03-12
issue #238 proposed fix

quetz:
Not sure if it's best way to solve my problem with websocket connections and gin, but it does it.


manucorporat:
Thanks for your fix, but I found to fix it without adding more variables.

PR #271, opened 2015-04-26
Fix TestBindingPOSTForm failing by adding Content-Type to request header...

j-e-k:
.... MIMEPOSTForm is required for req.ParseForm() to succeed in parsing the request body. Setting the Content-Type is only required for TestBindingPOSTForm, but I figured I might as well add it to all of the requests for consistency.


manucorporat:
yeah! I already fixed that. I did not expected that behaviour from the standard library.

PR #289, opened 2015-05-15
Append queue

saromanov:
Hi!
What do you think about simple queue to the context? We already have a functions
```go 
func (c *Context) Set(key string, value interface{})
```
and 
```go 
func (c *Context) Get(key string) (value interface{}, exists bool)
```
for the specified context, but this is just a queue for non-specified contexts.
This is looks like:

```go
func Foobar() gin.HandlerFunc {
    return func(c *gin.Context) {
        c.Enqueue("item1")
        c.Enqueue("item2")
    }
}

func main() {
    r := gin.New()
    r.Use(Foobar())

    r.GET("/test", func(c *gin.Context) {
        fmt.Println(c.Dequeue()) //item2
        fmt.Println(c.Dequeue()) //item1
    })
    r.Run(":8080")
}
```

### DIFF
--- a/context.go
+++ b/context.go
@@ -69,6 +69,7 @@ type Context struct {
 
 	Engine   *Engine
 	Keys     map[string]interface{}
+	Queue     []interface{}
 	Errors   errorMsgs
 	Accepted []string
 }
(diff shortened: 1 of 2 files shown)


manucorporat:
Well, that is a stack ( LIFO ).
can you provide a real world example? how people could use it?

saromanov:
Yes, Its my mistake.
I think i should close this PR now, because i found additional details in this feature, that i missed. Sorry, it was a haste.

manucorporat:
@saromanov don't worry!

PR #300, opened 2015-05-23
gin and pongo2 template engine example

nguyenthenguyen:
Hope this help someone want to use Gin with [pongo2](https://github.com/flosch/pongo2) like me!


manucorporat:
The pongo2 render should be in contrib.
https://github.com/gin-gonic/contrib

nguyenthenguyen:
OK. Thanks!

Issue #315, opened 2015-05-29
RunTLS does not work


njuettner:
The function RunTLS seems to be broken.  

```
package main

import (
	"fmt"
	"github.com/gin-gonic/gin"
	"net/http"
)

var sslkey string = ""
var sslcert string = ""

func main() {
	router := gin.Default()
	router.GET("/", func(c *gin.Context) {
		c.String(http.StatusOK, "hello world")
	})
	err := router.RunTLS("127.0.0.1:5443", sslcert, sslkey)
	if err != nil {
		fmt.Println("Server could not be started")
	}
}
```

After building, the application starts and listen

[GIN-debug] [WARNING] Running in DEBUG mode! Disable it before going production
[GIN-debug] GET   /                         --> main.func·001 (3 handlers)
[GIN-debug] Listening and serving HTTPS on 127.0.0.1:5443

I tried it also with a real certificate and a key but unfortunately it won't read them. Got the same message see below

```
curl https://127.0.0.1:5443
curl: (35) Unknown SSL protocol error in connection to 127.0.0.1:-9847
```

njuettner:
Instead of using RunTLS function trying to run http.ListenAndServeTLS directly with an empty key and certificate

```
package main

import (
	"github.com/gin-gonic/gin"
	"net/http"
)

var sslkey string = ""
var sslcert string = ""

func main() {
	router := gin.Default()
	router.GET("/", func(c *gin.Context) {
		c.String(http.StatusOK, "hello world")
	})
	err := http.ListenAndServeTLS("127.0.0.1:5443", sslcert, sslkey, router)
	if err != nil {
		fmt.Println("Could not start WebServer")
	}
}
```
I get the response I expected:

[GIN-debug] [WARNING] Running in DEBUG mode! Disable it before going production
[GIN-debug] GET   /                         --> main.func·001 (3 handlers)
Could not start WebServer


Using a certificate and key which exists in the directory
```
package main

import (
	"fmt"
	"github.com/gin-gonic/gin"
	"net/http"
)

var sslkey string = "./server.key"
var sslcert string = "./server.crt"

func main() {
	router := gin.Default()
	router.GET("/", func(c *gin.Context) {
		c.String(http.StatusOK, "hello world")
	})
	err := http.ListenAndServeTLS("127.0.0.1:5443", sslcert, sslkey, router)
	if err != nil {
		fmt.Println("Could not start WebServer")
	}
}
```

[...]

PR #318, opened 2015-05-29
Fix RunTLS

njuettner:


njuettner:
My Issue was related only to the develop branch, I didn't know that it was already commited to the master branch

https://github.com/gin-gonic/gin/issues/315

gin_test.go
```diff
@@ -148,3 +148,13 @@ func TestNoMethodWithGlobalHandlers(t *testing.T) {
...
 	assert.Equal(t, router.allNoMethod[2], middleware0)
 }
+
+func TestRunTLS(t *testing.T) {
+	certFile := ""
+	keyFile := ""
+	addr := ":5443"
+	router := New()
+
+	err := router.RunTLS(addr, certFile, keyFile)
```
manucorporat: this method blocks the calling goroutine. assert.NotNil() is never called.

PR #319, opened 2015-05-29
Fixing compile bug

bfosberry:
This causes a compile bug preventing godeps :(,  go version <1.4 I think (1.3.3)


manucorporat:
Thanks!

PR #343, opened 2015-06-15
Adding SPDY to RunTLS

itsmontoya:


itsmontoya:
@oneofone helped with this as well

slimmy:
I'm not sure defaulting to http2 is the preferred way to go yet. From what I've [read](https://github.com/bradfitz/http2/blob/master/README) the http2 package from Brad is eventually going to make it into the stdlib, maybe hold off until then? Or at least make it a new method like `RunTLSHTTP2` so that users have a choice?

manucorporat:
@itsmontoya http2 will eventually merge into the standard library, if it has not merged yet it probably means that it is not production ready.

Also, the gin package is a http router + tools, it is not a http server. ie. we should not mix different layers of abstraction.

Even though, you can still use http2 with Gin without modifying modifying the core.

```go
func main() {
	router := gin.Default()
	router.GET("/", index)
	// ....
	runHTTP2(router)
}

func runHTTP2(router *gin.Engine) {
	cfg := tls.Config{InsecureSkipVerify: false, RootCAs: x509.NewCertPool()}
	cfg.BuildNameToCertificate()

	srv := http.Server{
		Addr:           addr,
		Handler:        router,
		ReadTimeout:    5 * time.Second,
		WriteTimeout:   5 * time.Second,
		MaxHeaderBytes: 16384,
		TLSConfig:      &cfg,
	}

	http2.ConfigureServer(&srv, &http2.Server{})
	srv.ListenAndServeTLS(certFile, keyFile)
```

itsmontoya:
@manucorporat - RunTLS is already a part of the gin package so I figured it would be nice to have faster loads for SSL users. Would @slimmy's suggestion be acceptable?

manucorporat:
@itsmontoya 
>RunTLS is already a part of the gin package
it is shortcut for a standard production-ready code.

https://github.com/bradfitz/http2
```
This is a work-in-progress HTTP/2 implementation for Go.

It will eventually live in the Go standard library and won't require
any changes to your code to use.  It will just be automatic.

Status:

* The server support is pretty good. A few things are missing
  but are being worked on.
* The client work has just started but shares a lot of code
  is coming along much quicker.
```

thanks for the PR, but is not production ready, as I said, if anybody wants to use it, they should do it manually. Gin will not provide a shortcut for a in-development feature.

In addition, a method like `RunTLSHTTP2()` does not make sense, since http2 will eventually merge into `net/http`.

PR #348, opened 2015-06-23
Exporting routesInterface

WnP:
I like working with modules, with [routesInterface](https://github.com/gin-gonic/gin/blob/master/routergroup.go#L14) exported we can construct function that set routes for a give group or directly for the engine

PR #353, opened 2015-06-27
Rename gin.RoutesInterface to gin.RouterGroupInterface, and gin.routesInterface to gin.RoutesInterface

konjoot:
The reason of those changes is that till methods in the gin.routesInterface returns unexported type there are no ability to implement such interface outside of the gin package.

This PR relates to https://github.com/gin-gonic/gin/pull/348

### DIFF
--- a/gin.go
+++ b/gin.go
@@ -78,7 +78,7 @@ type (
 	}
 )
 
-var _ RoutesInterface = &Engine{}
+var _ RouterGroupInterface = &Engine{}
 
 // Returns a new blank Engine instance without any middleware attached.
 // The most basic configuration
@@ -154,7 +154,7 @@ func (engine *Engine) NoMethod(handlers ...HandlerFunc) {
 // Attachs a global middleware to the router. ie. the middlewares attached though Use() will be
 // included in the handlers chain for every single request. Even 404, 405, static files...
 // For example, this is the right place for a logger or error management middleware.
-func (engine *Engine) Use(middlewares ...HandlerFunc) routesInterface {
+func (engine *Engine) Use(middlewares ...HandlerFunc) RoutesInterface {
 	engine.RouterGroup.Use(middlewares...)
 	engine.rebuild404Handlers()
 	engine.rebuild405Handlers()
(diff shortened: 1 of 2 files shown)


manucorporat:
why do you want to "implement" the interface returned by the methods?

konjoot:
@manucorporat in our application there is main struct App, which looks something like:

```go
type App struct {
    Engine: *gin.Engine
}
```

So in tests sometimes very usefull to have the ability to mock App.Engine for speed and decoupling reasons, especially when we tests how our app interacts with gin. And there would be great if we could write something like that:

```go
type App struct{
    Engine: EngineIface
}

type EngineIface interface{
  // here we describes gin.Engine methods which we actually use in our code
}

// this gives us an ability to write a simple mock, which implements this interface:
type EngineMock struct {
   // dummy methods corresponding to EngineIface
}

// and finally in tests we can simply use our EngineMock instead of gin.Engine, when it is needed:
app := &App{Engine: &EngineMock{}}
```

Unfortunately we can't do that now because of the reason described above. I do not insist, but as for me, an exported interface that can be easily mocked outside of the package is extremely useful.

manucorporat:
right right, I agree! you already can! we already exported RoutesInterface, there is not need to export both interfaces. Since the exported interface includes all the method of the unexported interface.

manucorporat:
@konjoot just mock gin.Engine with `gin.RoutesInterface`

konjoot:
@manucorporat yes, gin.RoutesInterface is now exported, but signatures of methods in it, that  belongs to gin.routesInterface, have an non-exported type gin.routesInterface as the return value e.g. `POST(string, ...HandlerFunc) routesInterface`, so how can I implement corresponding mock method if `routersInterface` is non-exported - that's the issue. And that's why I made changes above. 

Perhaps there is a way around this, but I could not invent anything smarter than this PR.

manucorporat:
@konjoot ohh... I understand now. let me think about this.

konjoot:
closed because It is no longer relevant.

PR #359, opened 2015-07-02
Fix IsAborted() method

adwinsky:
According to the documentation IsAborted() method should return true if the current context was aborted. 

At the moment it  works only for the Handlers in which Abort() method was triggered. It doesn't work in the previous Handlers after the c.Next() call. It is because c.index will be higher then AbortIndex.

In my applications I would like to be able to determine whether the Context was aborted after the request was processed, without having to populate c.Errors slice.

### DIFF
--- a/context.go
+++ b/context.go
@@ -93,7 +93,7 @@ func (c *Context) Next() {
 
 // Returns if the currect context was aborted.
 func (c *Context) IsAborted() bool {
-	return c.index == AbortIndex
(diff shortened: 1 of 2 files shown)


manucorporat:
Can you rewrite the unit test with c.Next() instead of c.index++?

adwinsky:
Done. Hope you like this one better.

manucorporat:
@adwinsky thanks for the fix! good catch!

PR #363, opened 2015-07-04
RedirectTrailingSlash has no effect unless RedirectFixedPath is set

error10:
Which is not likely the desired behavior.

RedirectTrailingSlash setting is meant to cause a redirect, but the code is never called because the setting wasn't being checked. Instead RedirectFixedPath was being checked.


manucorporat:
thank you very much! I am going to add unit tests so this does not happen again

manucorporat:
@error10 https://github.com/gin-gonic/gin/commit/d4dec77afa9b8948dc09cbf7c8c057af24b48528

PR #375, opened 2015-07-12
add protobuf binding for gin

zhing:
We use gin as the server of our mobile app, choosing protobuf to Serialize our data for known reasons. So I add a module for protobuf binding. But I'm sorry to say we can not use 'binding:"len<10"' to validate on protobuf util now, this is what I'm doing on.


SlinSo:
I think you should change the mime type to a protobuf specific type.

most used types seem to be:
application/x-protobuf
application/protobuf

see https://issues.apache.org/jira/browse/HBASE-7472

zhing:
I'll have a try

zhing:
when I change the mime type to "application/x-protobuf"，the pull request build failed with go1.4.2. But It works well on my machine with go1.4.2, Why?

javierprovecho:
@zhing don't worry, ci failed because of a bad test behaviour.

javierprovecho:
@zhing thank you, merged in develop (91231d4801efb0db28cd21a4e7cd6e980a2183c8)

PR #381, opened 2015-07-22
Add option to skip logging specified endpoints

zaynetro:
**Possible use case:** 

Your application receives health check requests every minute and you don't want them to appear in the log.


javierprovecho:
@zaynetro thank you, merged in develop (ec30e4d727095ec138d481865a8862a717690467)

PR #386, opened 2015-07-25
Context.ClientIP(): only return the host's ip address

phicode:
http.Request.RemoteAddr contains the client's ip address and port number.
The reverse proxy headers X-Real-Ip and X-Forwarded-For usually only contain a client's ip address.
This change ensures that the ClientIP method behaves the same in both situations.

The additional logic results in one more allocation.
BenchmarkLoggerMiddleware is affected and goes from 14 to 15 allocations.


itsjamie:
I like..

PR #411, opened 2015-08-21
tests: make path assertions aware of vendoring

phicode:
The path of a package can change in a situation where
dependency vendoring is in use.
This change modifies gin's unit tests to allow such paths.


manucorporat:
@phicode thanks!

PR #416, opened 2015-08-27
Add SetCookie and GetCookie method for Context

se77en:


context.go
```diff
@@ -321,6 +322,64 @@ func (c *Context) Header(key, value string) {
...
+		}
+	}
+
+	c.Writer.Header().Add("Set-Cookie", cookie.String())
+}
+
+func (c *Context) GetCookie(name string) string {
+	cookie, err := c.Request.Cookie(name)
+	if err != nil {
+		return ""
```
ChristopherRabotin: Shouldn't this log the error instead of dying silently?

context.go
```diff
@@ -321,6 +322,64 @@ func (c *Context) Header(key, value string) {
 	}
 }
 
+func (c *Context) SetCookie(name string, value string, options ...interface{}) {
+	cookie := http.Cookie{}
+	cookie.Name = name
+	cookie.Value = url.QueryEscape(value)
+
+	if len(options) > 0 {
```
ChristopherRabotin: This sounds risky. Why just add the max age parameter in the SetCookie signature?

context.go
se77en: Getting cookie mustn't be a empty string, so I think logging an error is unnecessary

context.go
se77en: fix it

javierprovecho:
@se77en thank you, merged in develop (4892650ef8c095468cf3a1d8f166c28d45205653)

PR #423, opened 2015-09-06
Add  Engine.RunCustom method

snyh:
By this API the user can pass any net.Listener object to Engine.

It can be used by user's test code which run a Engine and bind to
a free port, then pass the real addres to the test function.

And I think if we added this API then the Engine.RunUnix is
unnecessary.

### DIFF
--- a/gin.go
+++ b/gin.go
@@ -225,6 +225,19 @@ func iterate(path, method string, routes RoutesInfo, root *node) RoutesInfo {
 	return routes
 }
 
+// RunCustom, run server by a net.Listener
+// the main usage is to test, by a custom net.Listener we can let os to choose a free
+// port and get it.
+//
+//     r := gin.Default()
+//     l, err := net.Listen(":0")
+//     l.Addr()
+//     r.RunCustom(l)
+func (engine *Engine) RunCustom(l net.Listener) error {
+	debugPrint("Listening and serving HTTP on %s\n", l.Addr())
+	return http.Serve(l, engine)
+}
+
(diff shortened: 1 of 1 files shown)


gin.go
```diff
@@ -225,6 +225,24 @@ func iterate(path, method string, routes RoutesInfo, root *node) RoutesInfo {
...
+port and get it.
+
+	r := gin.Default()
+	l, err := net.Listen(":0")
+	l.Addr()
+	r.RunCustom(l)
+*/
+func (engine *Engine) RunCustom(l net.Listener) error {
+	debugPrint("Listening and serving HTTP on %s\n", l.Addr())
+	if err := http.Serve(l, engine); err != nil {
```
mvader: Could be 

```go
return http.Serve(l, engine)
```

The error returned will be error or nil,  no need for an if.

gin.go
```diff
@@ -225,6 +225,24 @@ func iterate(path, method string, routes RoutesInfo, root *node) RoutesInfo {
 	return routes
 }
 
+/*
```
mvader: The convention in go is write comments with `//` for docs

snyh:
I have fix these problems, please review it.
But what cause the travis-ci failed?  https://travis-ci.org/gin-gonic/gin/builds/78979542 there hasn't any details inforamtion.

manucorporat:
I think this is a too low level feature and it makes code harder to understand, what does RunCustom() mean?

Just write:
```go
r := gin.Default()
listener, err := net.Listen(":0")
http.Serve(listener, r)
```

snyh:
It seems good.  :+1:

PR #441, opened 2015-09-21
Rawjson for the ability to write out a Buffer of raw JSON bytes when you have it

tvmaly:
please consider RawJSON  pull request.  I needed it for a case where I already had the JSON in a byte slice and I just wanted to write it out without incurring additional overhead costs of unmarshalling then re-marshalling it.  This is useful for cases such as redis with lua returning the raw JSON bytes of when you use a library like https://github.com/pquerna/ffjson where you need better performance for JSON.  

I also replaced the hardcoded test paths in the assertions so that it is easier for contributors to run the tests.

### DIFF
--- a/context.go
+++ b/context.go
@@ -5,6 +5,7 @@
 package gin
 
 import (
+	"bytes"
 	"errors"
 	"io"
 	"math"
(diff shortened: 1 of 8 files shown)


techjanitor:
It's not a context method but you can easily do this

	c.Writer.Header().Set("Content-Type", "application/json")
	c.Writer.Write(data)

manucorporat:
@techjanitor is right, or even better:
```go
c.Header("Content-Type", "application/json")
c.Data(data)
```

PR #458, opened 2015-10-09
Add c.PostFormExists(key)

RomanMinkin:
Method `c.PostFormExists(key)` returns `true` if field with given name exists in the form.

Right now there is no way to check if field exists in the form or not.
The functionality needed for example for `PATCH` requests when result of binding to a `string` gives false positive for field existence check.

Let's say we have struct for binding:
```go
type UserPayload struct {
	Gender    string `json:"gender" form:"gender"`
}

...

var userPayload UserPayload
c.Bind(&userPayload)
```

Now if I `POST/PATCH/PUT` both empty user object `{}` or user object with empty field `{gender: ""}` 
`userPayload.Gender` will be equal to an empty string. 

Hope my explanation is clear, but please feel free with any questions.
Thanks.

### DIFF
--- a/context.go
+++ b/context.go
@@ -198,6 +198,13 @@ func (c *Context) PostForm(key string) (va string) {
 	return
 }
 
+// PostFormExists returns `true` if field with given name exists in the form
+func (c *Context) PostFormExists(key string) (exists bool) {
+	_, exists = c.postForm(key)
+
+	return exists
+}
+
 // Param is a shortcut for c.Params.ByName(key)
 func (c *Context) Param(key string) string {
 	return c.Params.ByName(key)


RomanMinkin:
Forgot to mention that right now to check if field exists I need to do
```go
    if c.DefaultPostForm("gender", "missed") != "missed" {
}
```

RomanMinkin:
I can add tests if it can be considered for merging ;)

manucorporat:
@RomanMinkin 
how about this?
```go
if c.PostForm("gender") == "" {
 // no value
}
```

RomanMinkin:
@manucorporat in the case you described `c.PostForm("gender")` will always return an empty string either `gender` was not presented in a request at all or was presented with an empty value (`?gender=`).

Also public method `PostForm()` omits second parameter it gets form the private method `postForm()`
```go
// PostForm is a shortcut for c.Request.PostFormValue(key)
func (c *Context) PostForm(key string) (va string) {
	va, _ = c.postForm(key)
	return
}
```

Changing return statement of `PostForm()` would break the API tho...

So there are 2 options:
1. Change existing `PostForm()` and break the API
2. Add an addition method to expose functionality from private `postForm()`

p.s. thanks for replying and please let me know if it all makes sense ;)

manucorporat:
@RomanMinkin here's is what I think:

1. Forms in HTML are usually fixed, that means, there are a fixed number of inputs.
Let's say you have a HTML form with an email and a password input, even if the user leaves both fields empty, the browser will send a request like: `?email=&password=` i.e. the keys email and password will always exist!
A method like `c.PostFormExists()` would always return `true` in this use case, confusing API...
so what `c.PostFormExists()` means? in my opinion it's a confusing API.

```go
if c.PostFormExists("email") {
    db.FindUser(c.PostForm("email"))
}
```
^^ that code would not work as expected

While this one is perfect:
```go
if user := c.PostForm("email"); user != "" {
    db.FindUser(user)
}
```


So what if I really need to check if the key exists? then stick to the `net/http` API:
```go
	if values := req.PostForm["email"]; len(values) > 0 {
		    db.FindUser(values[0])
	}
```

it may be good to improve the documentation for `PostForm()`

RomanMinkin:
@manucorporat my bad that i did not specify my case, my apologies about that.

My use case is `PATCH` request with `Content-Type: multipart/form-data`, so I need partial update on resource and the only reason for `multipart/form-data` is because resource may contain binary file (media).

You are absolutely right about `?email=&password=` and it makes perfect sense for `POST` request, but in case with `PATCH` request with `?email=&` would represent setting `email` to empty string, otherwise there is no way to reset a certain resource field.

My understanding `PATCH`:
1. `email=mail@example.com` - set `email` to `"mail@example.com"`
2. `email=` - set `email` to empty string `""`
3. `email` not presented - do nothing with `email` filed

I'm just trying to lower the bandwidth and pressure on validator.

Agree on your `net/http` API solution, that's actually what private `postForm()` does i was just thinking to wrap it around in more friendly way ;)

p.s. btw I really love `gin`. Running it in production for relatively small API - just two servers behind LB each uses 4-40Mb of RAM, so far 1,5 months zero downtime ;)

manucorporat:
@RomanMinkin ok! now I totally understand the feature request! let me continue thinking about it.

yeah! `postForm()` has to do some fancy things in order to deal with the `net/http` mess.

manucorporat:
@RomanMinkin ok, here's what we are going to do:
we are going to expose the query() and postForm() methods, but we have to think about a good name.

RomanMinkin:
@manucorporat  Sounds good!

> "There are only two hard things in Computer Science: cache invalidation and naming things" (c)

And I'm better with **cache invalidation** ;)

manucorporat:
@RomanMinkin ^^genius hahaha

so... I think the Get prefix makes sense because:
- In Go it is not idiomatic to use Get to get something, but it is used across the standard library when (value, bool) is returned.
- Context already has a Get() method for user data, it returns (value, bool)
- Query() and GetQuery() have a similar naming because in fact they return the exact same thing haha
- Write a good documentation (comments) to explain this shit.
> "Comments are an apology for bad code."
Clean Code

RomanMinkin:
@manucorporat Sounds good to me. I see, it will align with context's `Get()` and `MustGet()` methods for sure. I'll update the PR  then.

manucorporat:
@RomanMinkin don't worry, I already have a local version... I am updating the tests.

manucorporat:
@RomanMinkin fixed:
https://github.com/gin-gonic/gin/commit/44978ff780ce235a830f6be7cfde42d350b1484d
https://github.com/gin-gonic/gin/commit/afc499f30694d145f91daa789dc6c7100373c493

RomanMinkin:
@manucorporat Awesome! Thanks for such a blazing response!

PR #465, opened 2015-10-18
fix query param name

mamal72:
fix query parameter name


manucorporat:
@mamal72 thanks! but someone else already fixed it.

PR #470, opened 2015-10-27
Clarify context.Abort docs

bwiggs:
I took a stab at clarifying the `context.Abort` func. Seems there's some confusion for us newbies as to how it's supposed to be used.


manucorporat:
@bwiggs thanks! and sorry for the delay!

PR #480, opened 2015-11-08
add custom template engine

SCys:
I realized this method, because I need to use a more compact pongo2 engine.

other way fix it: #449

### DIFF
--- a/README.md
+++ b/README.md
@@ -459,6 +459,86 @@ func main() {
 ```
 
 
+#### Custom rendering
+
+Custom your template engine use SetCustomRenderFunc()
+
+```go
+package main
+
+import (
+	"github.com/flosch/pongo2"
+	"github.com/gin-gonic/gin"
+	"net/http"
+	"errors"
+)
+
+func CustomRenderFunc(w http.ResponseWriter, name string, obj interface{}) error {
+	var err error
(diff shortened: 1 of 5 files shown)


robvdl:
This is done quite a bit differently to how I made a pongo2 template renderer for gin a few months ago:

https://github.com/robvdl/pongo2gin

Note that I was able to use the regular c.HTML() function with Pongo2 templates and without actual modification to Gin, as Gin already seems to support custom template renderers right now, it's just not really documented from what I can see.

Your version requires calling a special function to render templates, c.CustomRender(), I am not too sure if I actually like this, I prefer to keep using c.HTML() rather than introduce a new function for rendering custom templates.

SCys:
hi @robvdl , i use it with a small template engine write by myself.

robvdl:
Sorry but that doesn't really answer my question.  You are suggesting a new pull request to be merged into Gin to add support for custom template renderers, but I just mentioned that Gin already has support for a custom renderers in it's current code base, it's just not very well documented yet from what I can so. So why add it again?  This seems like code duplication, would it not be better to document the existing behaviour?

manucorporat:
@robvdl is right, there is a well designed support for custom renders.
@SCys

PR #483, opened 2015-11-13
Simplify the BasicAuth implementation

dolmen:
Use a map instead of a slice to store the preprocessed Accounts table.


manucorporat:
@dolmen we have a chance to change this with the next auth middleware, but we can not change the current one, because it is not backcompatible...

I will leave this PR open though

dolmen:
The change is in the `authPairs` type which as the name shows is not exported. So I don't see the backward compatibility issue.

dolmen:
I just rebased this PR on top of the latest master.

thinkerou:
any progress?

thinkerou:
@dolmen please re-commit the pull request to `master` branch, thanks!

PR #494, opened 2015-12-10
Fix indent of routes debug prints

dolmen:
'DELETE' is 6 chars and breaks indent of the routes dump (only 5 chars were reserved for the HTTP method).


manucorporat:
@dolmen thanks!

PR #499, opened 2015-12-17
fix godoc comment

178inaba:
It was fixed comments godoc.
This is the three places of "Bind, DefaultQuery, Query" in the Context.


178inaba:
Because such error is out it will withdraw once.

PR #509, opened 2016-01-12
Enhance template module

henrylee2cn:
Add methods:
1. H.TrySet(k string, v interface{}) H
2. (*Engine).HTMLTemplateDelims(left, right string);
3. (*Engine).HTMLTemplateFuncs(funcMap template.FuncMap);

Add template package.

Effect:
Support render templates from multiple directories;
support nested template;
support add functions to template;
support modify the templates delimiter.

### DIFF
--- a/debug.go
+++ b/debug.go
@@ -6,8 +6,9 @@ package gin
 
 import (
 	"bytes"
-	"html/template"
 	"log"
+
+	"github.com/gin-gonic/gin/template"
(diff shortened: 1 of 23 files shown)


manucorporat:
Gin templating is pluggable, 
check out: https://github.com/gin-gonic/contrib/tree/master/renders/multitemplate

PR #529, opened 2016-02-05
Develop

wpajqz:


javierprovecho:
Thank you for proposing this, but we will merge develop soon using a non forward strategy to reduce commits to one. Because of that, I'm closing this issue.

PR #542, opened 2016-02-23
Add integration test using httptest

bigwheel:
This test sample is useful when gin user writes integration test.
Because httptest runs httpserver asyncrhronous and selects port number automatically if we want.


javierprovecho:
@bigwheel thank you!

PR #547, opened 2016-02-29
Update form_mapping#mapForm to support reflect.Ptr

mephux:
Added reflect for ptr types to support the go-swagger generated http clients.


thinkerou:
ref #1336 please close the issue @appleboy

PR #555, opened 2016-03-08
Mpuncel/fix godoc handlername

mpuncel:


javierprovecho:
@mpuncel lol, don't worry! we'll sync master with develop soon :wink:

PR #558, opened 2016-03-10
Added gin.LoggerDisk which logs to screen and a file

sleven:
Added a clone function of gin.Logger() called gin.LoggerDisk(LogFile io.Writer) which accepts an io.Writer parameter to write to. LoggerDisk() will write to os.Stdout like normal and also write to file. I also added a hostname lookup field which I think is useful. 

Example usage: 

```
var f io.Writer
var err error
logfilename := fmt.Sprint(APP_DIRECTORY, "/log/access.log")

if _, err = os.Stat(logfilename); os.IsExist(err) {
  f, err = os.Open(logfilename)
} else {
  f, err = os.Create(logfilename)
}

router.Use(gin.LoggerDisk(f))
```

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"time"
+	"net"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


slimmy:
No need for this I think. You can just create your own [io.MultiWriter](https://golang.org/pkg/io/#MultiWriter) and pass that to `gin.LoggerWithWriter` for the same functionality (excluding the new field).

PR #577, opened 2016-04-02
added a support of a "block" template feature from go1.6. fixed #567

### LINKED ISSUE #567: Unable to use go 1.6 "block" template feature
Hello there!

I am trying to use the new "block" features in the `html/template` package in go 1.6, but I cannot get it to work inside a gin application.

I filled a thorough description of the bug as well as example code in the repository at https://github.com/esantoro/gin-template-bug
#### TLDR

layout.html

```
<!doctype html>
<html>
  <head>
    <title>{{block "title" .}} DEFAULT TITLE {{end}}</title>

  </head>
  <body>
    <div id="container" style="border: 1px solid black;">
      {{ block "content" .}}  DEFAULT CONTENT {{ end }}
    </div>

    <div class="footer" style="border: 1px solid yellow;">
      {{ block "footer" .}}
      DEFAULT FOOTER
      {{ end }}
    </div>
  </body>
</html>
```

index.html : 

```
{{define "title"}} CUSTOMIZED TITLE {{end}}
{{define "content"}} CUSTOMIZED CONTENT {{end}}
{{ define "footer" }} CUSTOMIZED FOOTER {{end}}
```

**Expected result**: Rendering index.html should fill layout.html with values from index.html

**Actual result**: index.html is rendered to a blank page.

im7mortal:
I added support of a ["block" template feature from go1.6](https://golang.org/doc/go1.6#template)
It relates with #567

I decided to use custom names for templates. It's pretty comfy. Check my example 

>_examples/template-blocks/main.go_

For a storing of templates I use map. It's only for reading. [It' safe](http://stackoverflow.com/questions/11063473/map-with-concurrent-access/11063566#11063566).


im7mortal:
@javierprovecho Can I have some callback? 
That feature is important for me. 
You can see #567 and #583

# What is issue?

You can't use "block" template feature from go1.6.  Check that [changes from go1.6](https://golang.org/doc/go1.6#template). It allow use different sub templates for one template like Jade style.

I don't know why the team didn't do it via one template....  Just check [official example](https://github.com/golang/example/blob/master/template/main.go)

They create separate template for every "block" sub template. It works only that way.

Gin use only one "html/template".Template.  You can set it via router.SetHTMLTemplate(template).
If you will  try add more templates it will give this warning

```
[WARNING] Since SetHTMLTemplate() is NOT thread-safe. It should only be called
at initialization. ie. before any route is registered or the router is listening in a socket:

    router := gin.Default()
    router.SetHTMLTemplate(template) // << good place
```

Also It doesn't work. You can't redefine default template. 

Check my [comment](https://github.com/gin-gonic/gin/issues/567#issuecomment-200802064)

# Why I think it has to be added in core of _gin_

Because that feature located in standard library. It's not some dialect or non standard feature.

# What I propose

Check my [comment](https://github.com/gin-gonic/gin/issues/567#issuecomment-204172555). 
Finally I used interfaces. And just added new render. I think will be comfy see it in source code.

# Tests

I added tests for render. But I couldn't do it for _AddHTMLTemplate_

I noticed that we don't have tests for templates. [Check it](https://github.com/gin-gonic/gin/blob/develop/gin_test.go#L14-L16)

im7mortal:
There are [a multitemplate render for gin](https://github.com/gin-gonic/contrib/tree/master/renders/multitemplate). Documentation is not clear about. I am going to add it.

PR #578, opened 2016-04-03
Rename test_helpers.go so it does not import httptest for users

divtxt:
The `test_helpers.go` file imports `net/http/httptest` which hooks into command line args.

Run any program using Gin with `-help`:
```
~/wgo/src/github.com/divtxt/lockd$ go run main.go -help
Usage of /var/folders/3q/77h08dhn0svg36nb4b53h1zr0000gp/T/go-build776975792/command-line-arguments/_obj/exe/main:
  -httptest.serve string
    	if non-empty, httptest.NewServer serves on this address and blocks
  -listen string
    	listen address (default ":2080")
exit status 2
```

Changing the file name to end with `_test.go` avoids the `-httptest.serve` flag showing up.

The downside of this fix is that this will result in `gin.CreateTestContext(..)` no longer being exported, but since it's not Gin's goal to provide test helpers, I hope that's ok.

### DIFF
--- a/test_helpers_test.go
+++ b/test_helpers_test.go
@@ -1,3 +1,4 @@
+// This file' name ends with "_test" to avoid pushing "httptest" import on users.
 package gin
 
 import (


javierprovecho:
@divtxt already merged in #581. I liked more that option: `test_helpers.go → helpers_test.go` instead of  `test_helpers.go → test_helpers_test.go`. Thank you.

divtxt:
That works too! 👍

PR #587, opened 2016-04-14
Write header immediately in AbortWithStatus()

roylou:
Otherwise, caller needs to invoke WriteHeaderNow himself after
AbortWithStatus(), which is error-prone.

Also modified ErrorLoggerT() such that it always writes log to response
body. Otherwise calling AbortWithStatus() will fail to write body because
c.Writer.Written() is set true by WriteHeaderNow().

It addresses #585


javierprovecho:
@roylou thank you!

PR #590, opened 2016-04-15
Default writers must be io.Writers

seregayoga:
Quote from docs
> DefaultWriter is the default io.Writer

But DefaultWriter has type `*os.File`, so i can't run code like this:
```
package main

import (
    "github.com/gin-gonic/gin"
    "io/ioutil"
)

func main() {
    gin.DefaultWriter = ioutil.Discard
}
```

```
$ go run main.go 
# command-line-arguments
./main.go:9: cannot use ioutil.Discard (type io.Writer) as type *os.File in assignment: need type assertion
```

It broke in [this](https://github.com/gin-gonic/gin/commit/9e930b9bdd5a29bac38fb491ffac4e7ea84b825d#diff-fd98eaa389d204eaf86f6913742c7551L34) commit

### DIFF
--- a/mode.go
+++ b/mode.go
@@ -5,6 +5,7 @@
 package gin
 
 import (
+	"io"
 	"os"
 
 	"github.com/gin-gonic/gin/binding"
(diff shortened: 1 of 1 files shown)


javierprovecho:
@seregayoga thank you, I didn't noticed, sorry about that.

PR #599, opened 2016-04-21
Improve BindJSON by avoiding panic when request.Body is nil

Focinfi:
1. If request body is nil, decoder.Decode will panic.
2. Returns the message as a error.
3. With testing.

### DIFF
--- a/binding/json.go
+++ b/binding/json.go
@@ -6,6 +6,7 @@ package binding
 
 import (
 	"encoding/json"
+	"errors"
 
 	"net/http"
 )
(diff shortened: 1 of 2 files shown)


Focinfi:
This PR is not aim at improving the coverage of testing, but when I test my

slimmy:
I wonder how this impacts performance.

I have APIs serving ~500 req/s with gin (using `BindJSON()`) and the thought of deferring 500 function-calls a second isn't appealing when we already have panic recovery in the framework.

Is it not possible to test for a nil body before calling `decoder.Decode`?

Focinfi:
I will a find a way to detect the nil body rather than defer in the `Bind()` function, cause performance is most important concern of `gin`

Focinfi:
@slimmy Actually testing for a nil body is nil, I reset the commit and avoid the panic in the simplest way without affecting the performance.

stxml:
This change is not necessary. The docs for http.Request say `Body` on a **server** request is always non-nil. Your test creates a **client** request and inserts it into the context.

> To create a request for use with testing a Server Handler use either ReadRequest or manually update the Request fields. See the Request type's documentation for the difference between inbound and outbound request fields.
-- from http.NewRequest docs

thinkerou:
@Focinfi please re-commit the pull request to `master` branch, thanks!

PR #620, opened 2016-05-17
Update README.md

betahu:
add code  to c.String


manucorporat:
@betahu Thank you very much!

PR #622, opened 2016-05-19
add Engine.ResetRoutes() for clean existed routes

tsaikd:
Signed-off-by: tsaikd tsaikd@gmail.com

### DIFF
--- a/gin.go
+++ b/gin.go
@@ -283,6 +283,11 @@ func iterate(path, method string, routes RoutesInfo, root *node) RoutesInfo {
 	return routes
 }
 
+// ResetRoutes reset routes in Engine, used when application need to change routes dynamically
+func (engine *Engine) ResetRoutes() {
+	engine.trees = make(methodTrees, 0, 9)
+}
+
 // Run attaches the router to a http.Server and starts listening and serving HTTP requests.
 // It is a shortcut for http.ListenAndServe(addr, router)
 // Note: this method will block the calling goroutine indefinitely unless an error happens.
--- a/gin_test.go
+++ b/gin_test.go
@@ -301,6 +301,21 @@ func TestAddRoute(t *testing.T) {
 	assert.Len(t, router.trees, 2)
 }
 
+func TestResetRoutes(t *testing.T) {
+	router := New()
+	router.addRoute("GET", "/", HandlersChain{func(_ *Context) {}})
+
+	assert.Len(t, router.trees, 1)
+	assert.NotNil(t, router.trees.get("GET"))
+	assert.Nil(t, router.trees.get("POST"))
+
+	router.ResetRoutes()
+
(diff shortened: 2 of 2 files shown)


thinkerou:
@tsaikd please re-commit the pull request to `master` branch, thanks!

thinkerou:
NO, please squash and recommit master branch not develop branch.

gin.go
```diff
@@ -281,6 +281,11 @@ func iterate(path, method string, routes RoutesInfo, root *node) RoutesInfo {
 	return routes
 }
 
+// ResetRoutes reset routes in Engine, used when application need to change routes dynamicly
```
dmarkham: dynamically

gin.go
```diff
@@ -281,6 +281,11 @@ func iterate(path, method string, routes RoutesInfo, root *node) RoutesInfo {
...
+func (engine *Engine) ResetRoutes() {
+	engine.trees = make(methodTrees, 0, 9)
+}
+
```
dmarkham: If the point of this is to change the router while it's running. Is this thread safe?

dmarkham:
`Engine.ResetRoutes()` Do we plan on this method being thread safe? This test fails the race detector currently. I did not test to see what it will do if you are hitting routes at a high rates and change it. 
 
If we choose to make it thread safe then I think it will snowball into `addRoute` also needing to be safe. I think the reason we are `ResetRoutes()` is to change the routes, and if the system is up and running while we are changing the routes what do we expect to happen?

```
func TestResetRoutes(t *testing.T) {
	router := New()
	router.addRoute("GET", "/", HandlersChain{func(_ *Context) {}})

	assert.Len(t, router.trees, 1)
	assert.NotNil(t, router.trees.get("GET"))
	assert.Nil(t, router.trees.get("POST"))
	var wg sync.WaitGroup
	wg.Add(1)
	go func() {
		router.ResetRoutes()
		wg.Done()
	}()
	wg.Add(1)
	go func() {
		router.ResetRoutes()
		wg.Done()
	}()
	wg.Wait()
	assert.Empty(t, router.trees)
	assert.Nil(t, router.trees.get("GET"))
	assert.Nil(t, router.trees.get("POST"))
}
```

tsaikd:
I use it for unit test only. If the behavior or use case is not what you expect, it's ok for me to close this PR.

dmarkham:
> I use it for unit test only. If the behavior or use case is not what you expect, it's ok for me to close this PR.

@tsaikd  Thank you for your use case! Wonder if there is a way can get this functionality for testing without making it a Public method?... What is the advantage of `ResetRoutes()` over just a new Router `router := gin.New()` in test cases?

tsaikd:
I forgot the detail about this. It's a few years ago PR. Seems there were no `gin.New()` (only global routing)?

PR #625, opened 2016-05-24
Write status code in external writer too

marvell:


HeyImAlex:
Are you wrapping the ResponseWriter with your own? Because I'm pretty sure Context.Writer and Context.writermem are the same object, one is just wrapped up in an interface.

mvader:
@heyimalex but if you change the `Context.Writer`, which is possible and middlewares like gin cache do, `Context.writermem` is not changed. And some things are only written in the private one, like the status https://github.com/gin-gonic/gin/blob/master/context.go#L378-L380

heyimalex:
If you change Context.Writer, don't you need to wrap the old one? If you wanted the wrapping-writers to get the status, then this method should *only* call Context.Writer.WriteHeader and not the one on writermem.

willnewrelic:
I'm with @heyimalex .  Shouldn't this function simply be:

```go
func (c *Context) Status(code int) {
	c.Writer.WriteHeader(code)
}
```

periclesroo:
Hey, any updates on this one?

thinkerou:
@marvell please re-commit the pull request to `master` branch, thanks!

willnewrelic:
@thinkerou @heyimalex @marvell 

I have opened a new PR against latest `master` with this change fixed:

https://github.com/gin-gonic/gin/pull/1606

PR #635, opened 2016-06-05
Replace uses of default/global log with package local instance.

### LINKED ISSUE #634: Don't zero out default log package flags.
I imported a package (github.com/mkevac/debugcharts) that has dependencies on gin. Suddenly all my log messages were missing the time stamps and I tracked it down to the [init function in debug.go](https://github.com/gin-gonic/gin/blob/develop/debug.go#L14)

rdwilliamson:
Fixes #634.

### DIFF
--- a/debug.go
+++ b/debug.go
@@ -8,11 +8,10 @@ import (
 	"bytes"
 	"html/template"
 	"log"
+	"os"
 )
 
-func init() {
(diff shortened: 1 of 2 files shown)


thinkerou:
please see #1560 and closing, thanks!

PR #636, opened 2016-06-07
Fixed Content-Type problem when `BindWith()` fails. Resolves #633.

### LINKED ISSUE #633: Content-Type is forced to text/plain when context.Bind() fails
If you call `context.Bind()` and it fails, `context.AbortWithError()` gets called, which means headers get sent. Result: you can't send a non-`text/plain` response later.
## Example

``` go
package main

import (
    "net/http"

    "github.com/gin-gonic/gin"
)

func main() {
    router := gin.Default()
    router.POST("/", func(c *gin.Context) {
        reqBody := make(map[string]string)
        err := c.BindJSON(&reqBody)
        if err != nil {
            c.JSON(http.StatusBadRequest, map[string]string{
                "error": err.Error(),
            })
        } else {
            c.JSON(http.StatusOK, map[string]string{
                "error": "No error. Everything is fine.",
            })
        }
    })
    router.Run(":8913")
}
```

If you send invalid JSON in the request, you should still expect the response to have the header `Content-Type: application/json; charset=utf-8`, but instead it has `Content-Type: text/plain; charset=utf-8`.

IMO `context.BindJSON()` should not call `context.AbortWithError()`. Instead, the caller should have to handle sending their own error response. If you agree, I'll certainly be happy to write the PR.

danslimmon:
Single-line fix! See #633.

### DIFF
--- a/context.go
+++ b/context.go
@@ -297,7 +297,7 @@ func (c *Context) BindJSON(obj interface{}) error {
 // See the binding package.
 func (c *Context) BindWith(obj interface{}, b binding.Binding) error {
 	if err := b.Bind(c.Request, obj); err != nil {
-		c.AbortWithError(400, err).SetType(ErrorTypeBind)
+		c.Error(err).SetType(ErrorTypeBind)
 		return err
 	}
 	return nil


javierprovecho:
closed in favor of a deprecation notice and two new methods for binding, see #855

PR #639, opened 2016-06-09
"/" (slash) as REST param does not work

jaceshim:
If we pass parameter to the REST function containing "/" (slash), then the param is not accepted, and results in an error. This seems to indicate that escaping special characters like "/" (slash) was not built in the gin framework. 

```
eg:- if we built an api like "/api/apps/{id}", then for a value of {id} as "/hello/world", it does not work. 
```

I added a Boolean ```UseRawPath``` to the ```struct Engine``` that if set to true, extracts the RawPath from the URL. The ```handleHTTPRequest``` is modified to include the check if ```UseRawPath``` is toggled or not.

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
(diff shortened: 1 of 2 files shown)


jaceshim:
i will again.

PR #640, opened 2016-06-09
"/" (slash) as REST param does not work

jaceshim:
If we pass parameter to the REST function containing "/" (slash), then the param is not accepted, and results in an error. This seems to indicate that escaping special characters like "/" (slash) was not built in the gin framework. 

```
eg:- if we built an api like "/api/apps/{id}", then for a value of {id} as "/hello/world", it does not work. 
```

I added a Boolean `UseRawPath` to the `struct Engine` that if set to true, extracts the RawPath from the URL. The `handleHTTPRequest` is modified to include the check if `UseRawPath` is toggled or not.

### DIFF
--- a/gin.go
+++ b/gin.go
@@ -78,6 +78,7 @@ type (
 		// handler.
 		HandleMethodNotAllowed bool
 		ForwardedByClientIP    bool
+		UseRawPath	       bool
 	}
 )
 
@@ -102,6 +103,7 @@ func New() *Engine {
 		HandleMethodNotAllowed: false,
 		ForwardedByClientIP:    true,
 		trees:                  make(methodTrees, 0, 9),
+		UseRawPath:		false,
 	}
 	engine.RouterGroup.engine = engine
 	engine.pool.New = func() interface{} {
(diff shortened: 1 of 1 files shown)


TheRealKira:
Wouldn't this work the same as `/api/app/*id`?

jaceshim:
@TheRealKira 
"/api/app/*id" It also occured 404 error.
Additionally when I try to encoding of client side then also occured 404 error.

TheRealKira:
@jaceshim that's interesting I tested this out with this code:
```go
package main

import (
    "github.com/gin-gonic/gin"
)

func main() {
    r := gin.Default()
    r.GET("/api/app/*id", func(c *gin.Context) {
        c.String(200, "working with ID:"+c.Param("id"))
    })
    r.Run()
}
```
and it looks like it works using the *wildcard param
```
[GIN] 2016/06/29 - 22:18:53 | 200 |      35.878µs | 127.0.0.1 |   GET     /api/app/hello/world
[GIN] 2016/06/29 - 22:18:56 | 200 |       8.519µs  | 127.0.0.1 |   GET     /api/app//hello/world
```

The only issue here is that the value stored in c.Param("id") has the leading "/" that needs to be trimmed from it.

and it seems to work for an "id" with the value of "/hello/world" and "//hello/world"

jaceshim:
@TheRealKira Thank you so much!
have enjoy contribute!

jaceshim:
i was defined REST API in main.go

e.g:-

```
router.GET("/api/apps/*appId/versions", function(c *gin.Context) {
  // bla bla...
})

router.GET("/api/apps/*appId", function(c *gin.Context) {
    // bla bla...
})

```
It does not work!

Do you have a Any Idea?

PR #701, opened 2016-09-09
Context.Get() does not need to test whether Keys is nil or not

maxatome:
Golang spec says: a nil map is equivalent to an empty map except that no elements may be added.

### DIFF
--- a/context.go
+++ b/context.go
@@ -166,9 +166,7 @@ func (c *Context) Set(key string, value interface{}) {
 // Get returns the value for the given key, ie: (value, true).
 // If the value does not exists it returns (nil, false)
 func (c *Context) Get(key string) (value interface{}, exists bool) {
-	if c.Keys != nil {
-		value, exists = c.Keys[key]
-	}
+	value, exists = c.Keys[key]
 	return
 }
 


maxatome:
This part of code is already covered by test TestContextSetGet. There is no need to add a new test. Even if I do, it would not change the coverage...
The coverage only decreased because the "if" statement has been removed, while it was already covered by TestContextSetGet. As the volume of covered code decreased, the coverage decreased...

javierprovecho:
@maxatome by default `c.Keys` is not initialized on context creation (https://github.com/gin-gonic/gin/blob/master/context.go#L64). Later when the first `c.Set(...)` is called (https://github.com/gin-gonic/gin/blob/master/context.go#L160), a map is initialized on demand. This is done like this to avoid an allocation.

@tboerger closing.

maxatome:
@javierprovecho you didn't understand my pull request. I know that c.Keys is nil on context creation and that it is initialized only when needed later.
Please re-read my first comment: the nil test is just not needed, that's all.
The code:
```
var Keys map[string]int
value := Keys["foo"]
```
is perfectly valid and do not generate any error at runtime -> no need to check whether Keys is nil or not.

javierprovecho:
@maxatome sorry, CI failed because you lighten up the code and coverage percent fell. Thanks for your pr.

@tboerger CI fail is ok, merge is secure, I already tested it.

PR #702, opened 2016-09-12
Fix context keys get set

rohan-j-shah:
Fix for : fatal error: concurrent map read and map write

### DIFF
--- a/context.go
+++ b/context.go
@@ -159,10 +159,10 @@ func (c *Context) Error(err error) *Error {
 // Set is used to store a new key/value pair exclusivelly for this context.
 // It also lazy initializes  c.Keys if it was not used previously.
 func (c *Context) Set(key string, value interface{}) {
+	c.KeysLocker.Lock()
 	if c.Keys == nil {
 		c.Keys = make(map[string]interface{})
 	}
-	c.KeysLocker.Lock()
 	c.Keys[key] = value
 	c.KeysLocker.Unlock()
 }


context.go
```diff
@@ -160,14 +162,18 @@ func (c *Context) Set(key string, value interface{}) {
 	if c.Keys == nil {
 		c.Keys = make(map[string]interface{})
 	}
+	c.KeysLocker.Lock()
```
cookingkode: I would suggest taking it before nil check to avoid a set being lost..

maxatome:
The locking mechanism involved in this pull request is not needed as the original gin.Context should not be used outside the original gin gonic goroutine without copying it (Copy method). In fact, it even hides some biggest errors!
If you have a "concurrent map read and map write" error, it means that you did a mistake. Probably using the same context in different goroutines without copying it.
If you need to share a gin.Context between several goroutines of your own, you first have to copy (Copy method) it then handle the locking mechanism yourself at your level, perhaps by composing the gin.Context in your own struct.
The problem with your patch is that it hides completely the fact that gin.Context instances are reused once the request is handled at the gin gonic level.

tsirolnik:
@maxatome I'm not sharing the context but I still do get panics sometimes even though the key was set.

See this issue #836

thinkerou:
@rohan-j-shah please re-commit the pull request to `master` branch, thanks!

thinkerou:
closing, ref #1391

PR #717, opened 2016-10-13
Remove net/http optional import from README

kris-nova:
Just throwing this out there..

But I feel like this step might be causing more confusion that it is helping. In this approach I think it's better to **under** prescribe than to **over** prescribe.

If the user is getting errors about a missing library, that is expected behavior, if a user is getting errors about an unused library, that is noise.

Just my opinion.. So I figured I would throw the PR out there..


tboerger:
Beside that you are targeting the wrong branch, prs are only accepted to develop branch.

Issue #469, opened 2015-10-26
How about change "golang.org/x/net/context" to "github.com/golang/net/context"


leedstyh:
Sorry to bother you, but golang.org is blocked in China. It's hard to `go get` code here.

So how about change "golang.org/x/net/context" to "github.com/golang/net/context"

muratsplat:
@leedstyh  You mean China government  banned  to access golang.org?

leedstyh:
@muratsplat Yeah, it's banned!

AlekSi:
Sorry, this will not work. Official import path is `golang.org/x/net/context`: https://github.com/golang/net/blob/master/context/context.go#L37

Issue #684, opened 2016-08-14
should change package context import directory when go 1.7


alex023:
in context.go ,package context is from "golang.org/x/net"

``` code
import"golang.org/x/net/context"
```

hope  can change to:

``` code
import "context"
```

when go 1.7 release. 
If can change now, better. Thanks.

stxml:
That would break gin for everybody who doesn't have 1.7 already, and it's not even necessary. That import is only used for the interface check and that hasn't changed.

gdiazlo:
Keeping up with the latest stable is advisable I think, old repos might disappear or worst, could become incompatible is subtle ways.

PR #724, opened 2016-10-19
Move golang.org/x/net/context.Context interface implementation check to tests

chiffa-org:
The only reason to import "golang.org/x/net/context" package is to check that `*gin.Context` implements `context.Context` interface. But this implies fetching whole "golang.org/x/net" package tree using `go get`. So moving this check to `context_test.go` is reasonable.
This also resolves issues #469 and #684.

### DIFF
--- a/context.go
+++ b/context.go
@@ -17,7 +17,6 @@ import (
 	"github.com/gin-gonic/gin/binding"
 	"github.com/gin-gonic/gin/render"
 	"github.com/manucorporat/sse"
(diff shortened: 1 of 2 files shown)


tboerger:
Maybe an exception for the newest go version? Because this got context integrated into the stdlib.

javierprovecho:
@tboerger for now we are supporting lower versions so the external import must stay. But in the near future or for gin.next we may drop support for older golang versions at once. Do you agree?

PR #733, opened 2016-11-08
Changed imports to gopkg instead of github in README

pmoncadaisla:
Replace old github.com imports with gopkg.in with fixed version


javierprovecho:
@pmoncadaisla thanks!

PR #849, opened 2017-03-26
Fix README

yhirano55:
before:

image

after:


yhirano55:
Oops, This problem has already been resolved on #837.

PR #898, opened 2017-04-24
Fix title and badge rendering

brian-brazil:


brian-brazil:
I see #897 already covers this.

PR #910, opened 2017-05-09
panic if err is nil on c.Error

ammario:
A panic here provides a more informative stack trace than the panic which would otherwise occur while errors are being collected.


nazwa:
The question is - should a nil error even cause a panic? Is that really a critical failure or is this something up to you to decide in your error handler... 🙄😱 🤔

ammario:
Calls to `c.Errors.Errors()` or `c.Errors.JSON()` will panic if an underlying error is nil due to `Error.Error` not checking if the underlying err is nil. Returning some kind of `ok` variable would make the API less ergonomic. Returning a string like "nil error" may be a decent idea but it wouldn't be easy to see where a nil error was set.

I don't think the panic is too big of a deal considering most people are using gin with some kind of recovery middleware.

kylecarbs:
@javierprovecho did you mean to merge this into master?

javierprovecho:
@kylecarbs it was an auto commit :wink:,

we are switching pr base branch from develop to master. develop will be removed in the future, master will be were the latest version is available, versions will be available through tags. I'll correct all the prs pointing to develop in the next days.

javierprovecho:
thanks @ammario

PR #924, opened 2017-05-28
Gitignore IntelliJ IDEA files.

hsluoyz:


tboerger:
You should put that into a global gitignore file, a project should not care about this kind of directory.

PR #974, opened 2017-07-04
improved swap

thinkerou:


thinkerou:
I think it should OK, so reopen.

PR #997, opened 2017-07-09
resolve #368 add option to UseNumber() in gin.Context.BindJSON()

### LINKED ISSUE #368: add option to UseNumber() in gin.Context.BindJSON()
When receiving a json object with number values and calling `context.BindJSON()` (or `context.Bind()`), the numbers will be unmarshalled to `float64` values. If needed, this can be switched to unmarshalling into `int` by using `json.Decoder.UseNumber()`. However, `BindJSON()` doesn't give this control. This should be added somehow:
- maybe a global switch, though this doesn't give per-request control.
- maybe an additional parameter or `...` parameter
- any other idea?

whirosan:
resolve #368  
I added an option EnableJsonDecoderUseNumber, so that you can choose whether or not to enable UseNumber in the JSON decoder.
Please review.

### DIFF
--- a/binding/json.go
+++ b/binding/json.go
@@ -10,7 +10,10 @@ import (
 	"github.com/json-iterator/go"
 )
 
-var json = jsoniter.ConfigCompatibleWithStandardLibrary
+var (
+	json                   = jsoniter.ConfigCompatibleWithStandardLibrary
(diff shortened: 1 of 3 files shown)


thinkerou:
Sorry I don't understand the pr to solve what?

PR #1003, opened 2017-07-11
update json-iterator revison to fix TestRenderIndentedJSON failed

easonlin404:
test failed, see https://github.com/json-iterator/go/issues/126

- [TestRenderIndentedJSON](https://github.com/gin-gonic/gin/blob/master/render/render_test.go#L55)


javierprovecho:
@easonlin404 nice catch! 😉

PR #1020, opened 2017-07-14
Fix context.Copy() race condition

furmmon:
This PR aims to point an existing race condition with the gin.Context and propose a solution.

The gin.Context stores keys in a `map[string]interface{}`. The function gin.Context.Copy()  (context.go:79) is supposed to give a thread safe copy of the initial gin.Context.

Instead of making a proper copy of the context.Keys, it copies the references of the map. This creates a race condition whenever the map context.Keys is accessed by concurrent go-routines.

With the following script you will get the raise condition instantly. The fix in the PR will solve it.
```go
package main

import (
        "github.com/gin-gonic/gin"
        "math/rand"
)

func main() {

        r := gin.Default()
        r.GET("/", func(ctx *gin.Context) {
                ctx.Set("1", 0)
                ctx.Set("2", 0)

                // Sending a copy of the gin.Context to two separate routines
                go ReadWriteKeys(ctx.Copy())
                go ReadWriteKeys(ctx.Copy())
        })
        r.Run(":8080")
}

func ReadWriteKeys(ctx *gin.Context) {
        for {
                ctx.Set("1", rand.Int())
                ctx.Set("2", ctx.Value("1"))
        }
}
```


javierprovecho:
@furmmon can you add a test at `context_test.go`, similar to your example code, or just checking that the reference of the map from the copied context is different from the original?

furmmon:
Sure, couldn't find a simple test for that at first. This test fails on the current master

gbbr:
@javierprovecho this seems like a fairly important thing to fix in a trivial set of changes - any plans on getting it in?

furmmon:
hey @appleboy, this is still an existing problem with the gin-gonic library, could you take a look at it
if you test the script above and `curl localhost:8080` you will see instantly the race condition

thinkerou:
@appleboy please review the pr, thanks!

PR #1031, opened 2017-07-17
use return not use else for reducing indent

thinkerou:


gin.go
```diff
@@ -156,19 +156,21 @@ func (engine *Engine) LoadHTMLGlob(pattern string) {
...
 		debugPrintLoadTemplate(template.Must(template.New("").Delims(engine.delims.Left, engine.delims.Right).Funcs(engine.FuncMap).ParseGlob(pattern)))
 		engine.HTMLRender = render.HTMLDebug{Glob: pattern, FuncMap: engine.FuncMap, Delims: engine.delims}
-	} else {
-		templ := template.Must(template.New("").Delims(engine.delims.Left, engine.delims.Right).Funcs(engine.FuncMap).ParseGlob(pattern))
-		engine.SetHTMLTemplate(templ)
+		return
 	}
+	templ := template.Must(template.New("").Delims(engine.delims.Left, engine.delims.Right).Funcs(engine.FuncMap).ParseGlob(pattern))
+	engine.SetHTMLTemplate(templ)
+	return
```
javierprovecho: just want to get an opinion, does adding a return on a case where is unneeded (no return values at all) makes code clearer for you to read? I somehow like both, depending on code complexity...

thinkerou: I usually add a return, but I read the official go lib and they not write return like this. So I will remove the last return for unifying gin's code style. Thanks!

PR #1064, opened 2017-08-13
fix(test): only check Location header

easonlin404:
https://github.com/gin-gonic/gin/pull/1063#issuecomment-322041800 fix testing failed. Please review it thanks.

### DIFF
--- a/routes_test.go
+++ b/routes_test.go
@@ -360,21 +360,21 @@ func TestRouterNotFound(t *testing.T) {
 		code   int
 		header string
 	}{
-		{"/path/", 301, "map[Location:[/path]]"},   // TSR -/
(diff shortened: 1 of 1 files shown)


routes_test.go
```diff
@@ -356,25 +356,25 @@ func TestRouterNotFound(t *testing.T) {
 	router.GET("/", func(c *Context) {})
 
 	testRoutes := []struct {
-		route  string
-		code   int
-		header string
+		route    string
+		code     int
+		Location string
```
thinkerou: Why is not `location`?

easonlin404: fixed. thanks for your reminder.

PR #1096, opened 2017-09-04
Fix colored logger on windows

Equim-chan:
In the previous version
```go
isTerm := true

if w, ok := out.(*os.File); !ok ||
	(os.Getenv("TERM") == "dumb" || (!isatty.IsTerminal(w.Fd()) && !isatty.IsCygwinTerminal(w.Fd()))) ||
	disableColor {
	isTerm = false
}
```

when `out` comes from `colorable.NewColorableStdout()` as [the doc suggested](https://github.com/gin-gonic/gin/blob/master/mode.go#L31), `out` will actually be `*colorable.Writer` type and will fail the type assertion `out.(*os.File)`, but `out` here, wrapped by `colorable`, is indeed a valid terminal that supports colors.

```go
package main

import (
	"fmt"
	"os"

	"github.com/mattn/go-colorable"
)

func main() {
	colored := colorable.NewColorableStdout()
	_, ok := colored.(*os.File)
	fmt.Println(ok) //=> false
}
```

Instead of judging whether `out` is an `io.Writer` wrapped by `colorable` or not in the case, I tried to utilize `github.com/mattn/go-colorable` directly here. Hopefully it's acceptable.

@mattn I think an `io.Writer` wrapped by `colorable` should be considered a terminal, but it may be hard to apply this to your `github.com/mattn/go-isatty`, as `github.com/mattn/go-colorable` depends on it and making the former depends on the latter may lead to circular reference.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -8,8 +8,10 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"runtime"
 	"time"
 
+	"github.com/mattn/go-colorable"
 	"github.com/mattn/go-isatty"
 )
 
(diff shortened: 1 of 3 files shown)


yosualb:
@Equim-chan since `gin.DefaultWriter = colorable.NewColorableStdout()` this error will still occur.

So the solution either:

1. add validation that gin.DefaultWriter is colorable.NewColorableStdout() or not, before terminal validation. (easier, but tight-dependency to github.com/mattn/go-colorable)
2. gin.DefaultWriter sets to immutable object. (hard, must check overall design)
3. add validation in github.com/mattn/go-isatty. (harder, because of other library)

thinkerou:
closing.

PR #1111, opened 2017-09-16
Using bytes.Buffer rather than + for string concat

atedja:


Equim-chan:
Can you provide a benchmark?

PR #1145, opened 2017-10-27
 use the Get/Set function of Header

thinkerou:


render/render.go
```diff
@@ -28,7 +28,7 @@ var (
 
 func writeContentType(w http.ResponseWriter, value []string) {
 	header := w.Header()
-	if val := header["Content-Type"]; len(val) == 0 {
-		header["Content-Type"] = value
+	if header.Get("Content-Type") == "" {
+		header.Set("Content-Type", value)
```
dalu: `header.Set("Content-Type", strings.join(value,""))`

but yeah, have it your way. assigning it directly is probably faster than going through .Set

thinkerou: Yes, I think that we should not use `Set`.

PR #1146, opened 2017-10-27
not use aliase when import package

thinkerou:


javierprovecho:
thank you @thinkerou 👍

PR #1170, opened 2017-11-13
307 is better than 301

qclaogui:
The browser will cache 301

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -984,7 +984,7 @@ Issuing a HTTP redirect is easy:
 
 ```go
 r.GET("/test", func(c *gin.Context) {
-	c.Redirect(http.StatusMovedPermanently, "http://www.google.com/")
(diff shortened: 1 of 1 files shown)


lutzhorn:
Why is 307 better than 301? It all depends on the scenario. If you want to indicate that the resource 'Moved Permanently', return 301. If you only want to trigger a 'Temporary Redirect', return 307.

I'd leave the `README.md` as it is. Everybody is free to use the status code defined in https://golang.org/pkg/net/http/#pkg-constants he sees fit.

scniro:
@lutzhorn I'd suggest `http.StatusTemporaryRedirect` would be a safer guardrail option to surface in the `README.md` example. I was testing this out just now and fell into the exact caching gotcha as @qclaogui. Now that I understand what happened, sure, I see your point. However, a 307 would have prevented the "issue" (me having a WTH moment) in the first place.

PR #1185, opened 2017-11-28
use lower if the struct is private

thinkerou:


dolmen:
I'll have to rebase my #483 which really improved auth.go.

PR #1196, opened 2017-12-10
use interface{} in map

noypi:
Prevents possible collision when using string keys, and uses package's data types.

Example usage:

```
package mypackage_with_keys

type _someKeyType int

const (
	CustomKey someKeyType = iota
	AnotherKey
)

func GetCustomData(c *Context) (...) {
	o, has := c.Get(CustomKey)
        .....
	return
}
```

### DIFF
--- a/context.go
+++ b/context.go
@@ -49,7 +49,7 @@ type Context struct {
 	engine *Engine
 
 	// Keys is a key/value pair exclusively for the context of each request.
-	Keys map[string]interface{}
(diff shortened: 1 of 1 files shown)


thinkerou:
@noypi sorry I not agree the pull request, use `int` index or other type should not framework do, it's my opinion.

thoeni:
So maybe this can be closed? 🤔

thinkerou:
closing. thanks!

PR #1238, opened 2018-01-28
add prefix from X-Forwarded-Prefix in redirectTrailingSlash

tudurom:
Very useful when using Gin with some reverse proxies like Traefik


thinkerou:
@tudurom please fix conflict, thanks!

gin.go
```diff
@@ -405,7 +406,13 @@ func serveError(c *Context, code int, defaultMessage []byte) {
...
 func redirectTrailingSlash(c *Context) {
 	req := c.Request
-	path := req.URL.Path
+	prefix := path.Clean(c.Request.Header.Get("X-Forwarded-Prefix"))
+	path := ""
+	if prefix == "." {
+		path = req.URL.Path
+	} else {
+		path = prefix + "/" + req.URL.Path
+	}
```
thinkerou: change to:
```go
path := req.URL.Path
if prefix := path.Clean(c.Request.Header.Get("X-Forwarded-Prefix")); prefix != "." {
    path = prefix + "/" + req.URL.Path
}
```
do you think?

PR #1253, opened 2018-02-16
Get Body From Context

davidbayo10:
Method for getting body request with Content-Type like MIMEJSON, MIMEPOSTForm or MIMEMultipartPOSTForm


davidbayo10:
Why the build for golang version 1.9 failed? 😔

PR #1268, opened 2018-03-04
feat(context): formating json response

bruinxs:
if cant add feature like this, so i can have a standard JSON format response to caller


thinkerou:
Unnecessary, closing.

PR #1271, opened 2018-03-07
Fix intercepting headers in middlewares

tjamet:
As explained in the TestInterceptedHeader test, in case a middleware
filters out the headers, this middleware can be done inefficient in case
one following handler is using c.String or other methods writing to the
response body directly.

This commit fixes the issue by using c.Writer when writing the Status as
done in other c.Header, c.SetCookie and other response writers.

The bug has been originally discovered using
https://github.com/gin-contrib/gzip where a failing test has been added
here: https://github.com/tjamet/gzip/blob/header/gzip_test.go#L71

Signed-off-by: Thibault Jamet <tjamet@users.noreply.github.com>


tjamet:
@manucorporat having a look at the history, you seems to have the most knowledge on `context.go`, may I ask you for help reviewing this change?
I am experiencing difficulties using the gzipping middleware together with the context `Status` helper, gzip fails to filter out the `Content-Length` header

boindil:
is there a reason for this not being closed?

the adjusted line https://github.com/gin-gonic/gin/blob/f1e942889abdab773538f215ec6aa242f994c6d0/context.go#L828 was already changed in https://github.com/gin-gonic/gin/pull/1606

thinkerou (approved):

context_go17_test.go
```diff
@@ -0,0 +1,50 @@
...
+type interceptedWriter struct {
+	ResponseWriter
+	b *bytes.Buffer
+}
+
+func (i interceptedWriter) WriteHeader(code int) {
+	i.Header().Del("X-Test")
+	i.ResponseWriter.WriteHeader(code)
+}
+func TestInterceptedHeader(t *testing.T) {
```
thinkerou, line 23: please add it to `context_go17_test.go` which exist, thanks!

PR #1349, opened 2018-05-09
drop dependence on golang.org/x/net/context.Context

alex023:
now is 2018,  go1.10 released. May be, it does not necessary for checking the implementation of the x/net/context.Context interface. Just stand lib is enough.


thinkerou:
No, need to consider `go1.6` `go1.7` user.

PR #1419, opened 2018-07-02
Use camel case instead of ALL_CAPS

crispgm:
ENV_GIN_MODE -> EnvGinMode


thinkerou:
it will break forward compatibility.

crispgm:
@thinkerou well, should I close?

thinkerou:
@crispgm sorry, I understand error, it not break, because set environment var using `GIN_MODE`, so not close, need cr, thanks!

thinkerou:
@crispgm maybe we should use `ALL_CAPS` for `const`, https://github.com/Unknwon/go-code-convention/blob/master/en-US/naming_rules.md#constants, but I not find the official document.

crispgm:
@thinkerou Nope, mixed caps is the common practice for constants, which is used by go linters.
See <https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps>.

PR #1482, opened 2018-08-13
chore: use http.Status* instead of hard code

thinkerou:


thinkerou:
Oh, sorry @appleboy still have one error with go1.6.

PR #1487, opened 2018-08-14
Set default time format in form binding

lokhman:
To make things working in line with `encoding/json` decoder, I'd suggest to set default time format as `time.RFC3339`.

```
type input struct {
	Field time.Time `form:"field" json:"time"`
}
```

This will make values bound in the same way as if you are using structures as above and different bindings depending on the request type.


thinkerou:
@lokhman thanks, LGTM

PR #1530, opened 2018-09-04
Update: examples/graceful-shutdown/server.go

ffhelicopter:
It's necessary that  catching  ctx.Done()
        
	if err := srv.Shutdown(ctx); err != nil {
		log.Fatal("Server Shutdown:", err)
	}

	// catching ctx.Done(). timeout of 5 seconds.
	select {
	case <-ctx.Done():
	}

	log.Println("Server exiting")

- With pull requests:


thinkerou:
please run `go fmt` to format code.

examples/graceful-shutdown/graceful-shutdown/server.go
```diff
@@ -1,4 +1,4 @@
-// +build go1.8
+// +build go1.11
```
thinkerou: why change to `go1.11`?

ffhelicopter: I am using  v1.11. If it is not suitable， I will change it back to v1.8.

thinkerou: yes, you should not update.

examples/graceful-shutdown/graceful-shutdown/server.go
```diff
@@ -44,5 +44,10 @@ func main() {
 	if err := srv.Shutdown(ctx); err != nil {
 		log.Fatal("Server Shutdown:", err)
 	}
+	// catching ctx.Done(). timeout of 5 seconds.
```
thinkerou: you also should update readme example, here: https://github.com/gin-gonic/gin#graceful-restart-or-stop
because ci run `make embedmd` failed.

thinkerou:
`readme.md` should be the same as `server.go`, the error info:

```
0.03s$ make embedmd
embedmd -d *.md
@@ -1590,6 +1590,11 @@
 	if err := srv.Shutdown(ctx); err != nil {
 		log.Fatal("Server Shutdown:", err)
 	}
+	// catching ctx.Done(). timeout of 5 seconds.
+	select {
+	case <-ctx.Done():
+		log.Println("timeout of 5 seconds.")
+	}
 	log.Println("Server exiting")
 }
 ```
make: *** [embedmd] Error 2
```

iuouiyiuty:
@ffhelicopter @thinkerou 
I think there is unnecessary and shuldn't to add catching ctx.Done(). Because it was catching inside `srv.Shutdown() `.
Besides 
```
// catching ctx.Done(). timeout of 5 seconds.
select {
    case <-ctx.Done():
}
```
Running this code will always block until timeout. just remove it

PR #1532, opened 2018-09-05
readme: remove console image and add fold button

thinkerou:
- remove console image on the top of readme, I think it's not important, but need review.
- add fold button which reduce readme document length, it will fold some less important content.

like this:

image


thinkerou:
closing, building website.

PR #1112, opened 2017-09-17
Support negotiation wildcards, fix #391

Equim-chan:
Support negotiation wildcards, fix #391.

PR #1535, opened 2018-09-08
Updated content-type parsing to respect wildcards.

cjyar:
The existing content-type negotiation doesn't handle wildcard types like */*, text/*, etc. This PR adds that support.


thinkerou:
@cjyar the same as #1112

thinkerou:
@cjyar thank your pull request, I will close it, if you have any idea, welcome to commit issue/pr.

PR #1563, opened 2018-09-21
feat:add func BeforeHandle to Engine

RedDragonet:
In my project, API version control like [github API rule](https://developer.github.com/v3/media/#request-specific-version):
```bash
curl https://api.github.com/users/technoweenie -I \
 -H "Accept: application/vnd.github.v3.full+json"
```

So i change `Engine` struct, added `BeforeHandle` function, support like this complex route match.
And i think this function will more usage scenarios.

> Basic usage demo:

```golang
func v1Foo(c *gin.Context) {
	c.String(http.StatusOK, "v1")
}
func v2Foo(c *gin.Context) {
	c.String(http.StatusOK, "v2")
}

func githubVersion(r *gin.Engine) {
	r.BeforeHandle = func(c *gin.Context, httpMethod, path, rawPath *string) {
		accept := c.Request.Header.Get("Accept")
		reg := regexp.MustCompile(`application\/vnd\.github\.v(\d*?)\+json`)
		res := reg.FindStringSubmatch(accept)

		if 2 == len(res) {
			*path = "/v" + res[1] + (*path)
		}
	}
}
func main() {
	r := gin.Default()
	githubVersion(r)

	r.Handle(http.MethodGet, "/v1/foo", v1Foo)
	r.Handle(http.MethodGet, "/v2/foo", v2Foo)
	r.Run()
} 
```

> run
```bash
curl -X GET \
  http://localhost:8080/foo \
  -H 'Accept: application/vnd.github.v2+json' \
```

### DIFF
--- a/gin.go
+++ b/gin.go
@@ -104,6 +104,11 @@ type Engine struct {
 	noMethod         HandlersChain
 	pool             sync.Pool
 	trees            methodTrees
+
+	// If defiend, invoke before `handleHTTPRequest`
+	// Change `httpMethod`, `path`, `rawPath` for complex route match,it will not change origin `Context.Request`
+	// Note: If necessary, DON'T MODIFIY `Context`
+	BeforeHandle func(c *Context, httpMethod, path, rawPath *string)
 }
 
 var _ IRouter = &Engine{}
@@ -342,9 +347,15 @@ func (engine *Engine) HandleContext(c *Context) {
 func (engine *Engine) handleHTTPRequest(c *Context) {
 	httpMethod := c.Request.Method
 	path := c.Request.URL.Path
+	rawPath := c.Request.URL.RawPath
+	if engine.BeforeHandle != nil {
+		engine.BeforeHandle(c, &httpMethod, &path, &rawPath)
+	}
+
 	unescape := false
-	if engine.UseRawPath && len(c.Request.URL.RawPath) > 0 {
-		path = c.Request.URL.RawPath
+
+	if engine.UseRawPath && len(rawPath) > 0 {
+		path = rawPath
 		unescape = engine.UnescapePathValues
 	}
 


chainhelen:
IMO, [custom-middleware](https://github.com/gin-gonic/gin#custom-middleware) and [redirects(Router redirect)](https://github.com/gin-gonic/gin#redirects) are enough.

RedDragonet:
> IMO, [custom-middleware](https://github.com/gin-gonic/gin#custom-middleware) and [redirects(Router redirect)](https://github.com/gin-gonic/gin#redirects) are enough.

@chainhelen 
no, both change `c.Request.URL.Path`, origin request uri will be changed

chainhelen:
IMO，even custom-middleware is better than your BeforeHandle.You also save var by `ctx.Set`.

Bussiness should do it itself.The `BeforeHandle` is too obscure(just like middleware but with route).  

But this is just my personal suggestion, wait for administrator of gin.

RedDragonet:
thanks @chainhelen , this is good suggestion. 
And waiting for reply from administrator of gin.

thinkerou:
@RedDragonet sorry, I don't know what's your question mean?

```go
➜  ~ curl -v 'http://localhost:8080/rest/n/api' -H 'Accept:application/vnd.github.v2+json'
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> GET /rest/n/api HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.54.0
> Accept:application/vnd.github.v2+json
>
< HTTP/1.1 200 OK
< Date: Sat, 22 Sep 2018 01:00:03 GMT
< Content-Length: 0
<
* Connection #0 to host localhost left intact
```

RedDragonet:
Gin use request uri routing.Is there any way to use custom request header routing?
Just like
```bash
curl https://api.github.com/users/technoweenie -I \
 -H "Accept: application/vnd.github.v3.full+json"
```
routing to
```golang
func technoweenieV3(){}
```

PR #1600, opened 2018-10-17
Pass MaxMultipartMemory when FormFile is called

isgj:
When `gin.Context.FormFile("...")` is called the `engine.MaxMultipartMemory` is never used. This PR makes sure that the `MaxMultipartMemory` is passed and removes 2 calls to `http.Request.ParseForm` since they are called from `http.Request.ParseMultipartForm`

### DIFF
--- a/context.go
+++ b/context.go
@@ -414,7 +414,6 @@ func (c *Context) PostFormArray(key string) []string {
 // a boolean value whether at least one value exists for the given key.
 func (c *Context) GetPostFormArray(key string) ([]string, bool) {
 	req := c.Request
-	req.ParseForm()
 	req.ParseMultipartForm(c.engine.MaxMultipartMemory)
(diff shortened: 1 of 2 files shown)


thinkerou:
@isgj please add some test, thanks!

isgj:
There was already a test for the `happy path`. I wrote a test when it fails to check that it returns the error. I you think that another type of test  can be written just let me know.

context.go
```diff
@@ -414,7 +414,6 @@ func (c *Context) PostFormArray(key string) []string {
 // a boolean value whether at least one value exists for the given key.
 func (c *Context) GetPostFormArray(key string) ([]string, bool) {
 	req := c.Request
-	req.ParseForm()
```
thinkerou: @isgj Why remove it? thanks!

thinkerou:
@appleboy please help review, thanks!

PR #1614, opened 2018-10-29
support ArrayMap for post form

farmerx:
- go version: 1.9
- gin version (or commit ref): master branch
- operating system: 

## Description
image

this image is my  http post params .
I'm going to use c.GetPostFormMap to parse processes field, but I didn't get the expected results.
So I wrote a solution myself, as follows：
```
// GetPostFormArrayMap returns a map for a given form key, plus a boolean value
// whether at least one value exists for the given key.
func (c *Context) GetPostFormArrayMap(key string) ([]map[string]string, bool) {
	req := c.Request
	req.ParseForm()
	req.ParseMultipartForm(c.engine.MaxMultipartMemory)
	dicts, exist := c.getArrayMap(req.PostForm, key)
	if !exist && req.MultipartForm != nil && req.MultipartForm.File != nil {
		dicts, exist = c.getArrayMap(req.MultipartForm.Value, key)
	}
	return dicts, exist
}

// get is an internal method and returns a map which satisfy conditions.
func (c *Context) getArrayMap(m map[string][]string, key string) ([]map[string]string, bool) {
	dicts := make(map[string]map[string]string)
	exist := false
	for k, v := range m {
		if i := strings.IndexByte(k, '['); i >= 1 && k[0:i] == key { // 确定name[1]key
			num := ``
			if j := strings.IndexByte(k[i+1:], ']'); j >= 1 {
				num = k[i+1:][:j]
			}
			if j := strings.IndexByte(k[i+1:], '['); j >= 1 {
				exist = true
				field := k[i+1:][j+1 : len(k[i+1:])-1]
				if dict, ok := dicts[num]; ok {
					dict[field] = v[0]
					dicts[num] = dict
				} else {
					dicts[num] = map[string]string{field: v[0]}
				}
			}
		}
	}
	arr := make([]map[string]string, 0)
	for _, value := range dicts {
		arr = append(arr, value)
	}
	return arr, exist
}
```
this image is my parse result .
image


farmerx:
测试用力写完了，但是找不到在哪里推了

context.go
```diff
@@ -452,6 +452,48 @@ func (c *Context) PostFormMap(key string) map[string]string {
...
+	req.ParseForm()
+	req.ParseMultipartForm(c.engine.MaxMultipartMemory)
+	dicts, exist := c.getArrayMap(req.PostForm, key)
+	if !exist && req.MultipartForm != nil && req.MultipartForm.File != nil {
+		dicts, exist = c.getArrayMap(req.MultipartForm.Value, key)
+	}
+	return dicts, exist
+}
+
+// get is an internal method and returns a map which satisfy conditions.
```
thinkerou: change `get` to `getArrayMap`

context.go
```diff
@@ -452,6 +452,48 @@ func (c *Context) PostFormMap(key string) map[string]string {
...
+func (c *Context) getArrayMap(m map[string][]string, key string) ([]map[string]string, bool) {
+	dicts := make(map[string]map[string]string)
+	exist := false
+	for k, v := range m {
+		if i := strings.IndexByte(k, '['); i >= 1 && k[0:i] == key { // 确定name[1]key
```
thinkerou: remove comment.

context.go
```diff
@@ -452,6 +452,48 @@ func (c *Context) PostFormMap(key string) map[string]string {
...
+			num := ``
```
thinkerou: why use num := `` syntax? I don't suggest.

dmarkham:
This feels like a step to far for a helper we provide use json! We have helpers to make things easy this is just too much  :-1:     Now that I got that out of the way :)

This is pretty old and I don't think the remaining changes are much if we are wanting to clean this up I'm guessing one of us will need to do the last few changes @appleboy  @thinkerou if @farmerx  does not clean  up the requested changes in the next day or so I'll make a new PR  from his work and get it done.

dmarkham:
@farmerx  I'm sorry if it sounded like I was asking you to close this (I was not). I was looking to see if you wanted to do the last few updates on this PR and if not I would do it if the core team was still interested in this PR.

PR #1637, opened 2018-11-10
cookie SameSite support

tuhochi:
- Support for SameSite cookie via bool because Lax or Strict has same result
Source: https://www.netsparker.com/blog/web-security/same-site-cookie-attribute-prevent-cross-site-request-forgery/


thinkerou:
duplicated #1615

PR #1220, opened 2018-01-03
router: handling file not found in StaticFS() to use Router's NoRoute

okyantoro:
Refs https://github.com/gin-gonic/gin/issues/1182

PR #1663, opened 2018-11-25
Router: Route StaticFS() not found to Router's NoRoute()

MetalBreaker:
Closes #1220


routergroup.go
```diff
@@ -185,11 +185,19 @@ func (group *RouterGroup) StaticFS(relativePath string, fs http.FileSystem) IRou
 func (group *RouterGroup) createStaticHandler(relativePath string, fs http.FileSystem) HandlerFunc {
 	absolutePath := group.calculateAbsolutePath(relativePath)
 	fileServer := http.StripPrefix(absolutePath, http.FileServer(fs))
-	_, nolisting := fs.(*onlyfilesFS)
+
```
thinkerou: why remove the line? can you explain? thanks!

MetalBreaker: I removed it because it wasn't required, I could directly use fs.Open() to get both if it exists (no need to use fs.(*onlyfilesFS)) and if it's accessible. Two function calls (half-reduntant) wouldn't really be efficient.

thinkerou: Ok, thanks!

MetalBreaker:
Made it automergeable now. Thanks for reviewing :)

Issue #1656, opened 2018-11-23
travisCI fail but not found


thinkerou:
fail info: https://travis-ci.org/gin-gonic/gin/jobs/458636581
pull request: #1387 

cc @BanzaiMan  thanks!

thinkerou:
@BanzaiMan Thank your reply! Sorry for not clear describe the question.

1. use the command `go test ...` to execute unit test

2. And it have some FAIL, like:

```
=== RUN   TestUriBinding
--- FAIL: TestUriBinding (0.00s)
    binding_test.go:690: 
        	Error Trace:	binding_test.go:690
        	Error:      	Not equal: 
        	            	expected: string("")
        	            	actual  : map[string]interface {}(map[string]interface {}(nil))
        	Test:       	TestUriBinding
```

3. But it show Success, like:

![image](https://user-images.githubusercontent.com/4488261/48997598-76c26680-f18b-11e8-9933-75ac5a8f979b.png)

I hope it show Fail, like:

![image](https://user-images.githubusercontent.com/4488261/48997650-9d809d00-f18b-11e8-92a5-b4ac3b877709.png)

Sorry, I don't know travis-ci can catch the case?

Maybe travis-ci have config which can solve it, but I'm not found it now.

Thanks again!

BanzaiMan:
The command you are running is `make test`, not any of the individual `go test` commands. If `make test` exits with zero, it is successful. If you want any of the `go test` commands to trip the wire and cause the failure, you will have to modify your Make target to accomplish it.

PR #1671, opened 2018-11-27
ci: break when test fail

### LINKED ISSUE #1670: How to break test when go test FAIL?
ref #1656 

Now run `make test` script:

```
test:
	echo "mode: count" > coverage.out
	for d in $(TESTFOLDER); do \
		$(GO) test -v -covermode=count -coverprofile=profile.out $$d; \
		if [ -f profile.out ]; then \
			cat profile.out | grep -v "mode:" >> coverage.out; \
			rm profile.out; \
		fi; \
	done
```

If use `$(GO) test -v -covermode=count -coverprofile=profile.out $$d; | grep "^--- FAIL" | exit 1;`, it can output log to console.

So, how should we?

thinkerou:
closes #1670 #1656 

please @appleboy review the scheme, and fail case have fixed #1669 , please first review and merge it, thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ vendor/*
 coverage.out
 count.out
 test
+profile.out
+tmp.out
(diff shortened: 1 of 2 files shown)


thinkerou:
#1682 need to merged firstly.

PR #1690, opened 2018-12-10
context: inherits context cancelation and deadline from http.Request context for Go>=1.7

rbeuque74:
*gin.Context implements standard context.Context methods, but always
returns data as context is still valid. Since Go 1.7, http.Request now
contains a context.Context object, which can be controlled by the
http.Server to indicates that the context is now closed, and persue of
request should be canceled.

This implements the propagation of http.Request context methods inside
gin.Context to have HTTP context cancelation information at gin.Context
level.

This is retro-compatible for Golang 1.6, and have no effect if http.Request have no context (as http.Request always provide context.Background is ctx is nil)

### DIFF
--- a/context_17_test.go
+++ b/context_17_test.go
@@ -7,9 +7,12 @@
 package gin
 
 import (
+	"bytes"
+	"context"
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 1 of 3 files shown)


rbeuque74:
Build failing because of Travis timeout, if one of the maintainer can restart the build :)
Thanks!

context_17_test.go
```diff
@@ -7,11 +7,14 @@
...
 
 import (
+	"bytes"
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
+	"golang.org/x/net/context"
```
thinkerou: why not use `context`?

rbeuque74: Because my IDE thought that it was the best fit when autocompleting the import...
Thanks, fixed.

PR #1691, opened 2018-12-11
feat: ShouldBindWith support for reading saved body bytes

myml:
ShouldBindWith support for reading saved body bytes.
use ShouldBindJSON,ShouldBindXML and ShouldBindYAML after ShouldBindBodyWith


thinkerou:
thank your pull request, we have had `ShouldBindBodyWith`, why need to add it? I don't think we need it.

myml:
> 谢谢你的拉取请求，我们有`ShouldBindBodyWith`，为什么需要添加它？我认为我们不需要它。

`ShouldBindWith` does not check for saved body bytes,I can only use `ShouldBindBodyWith` after using `ShouldBindBodyWith` once. `ShouldBindJSON`，`ShouldBindXML` and `ShouldBindYAML` complain EOF

myml:
> thank your pull request, we have had `ShouldBindBodyWith`, why need to add it? I don't think we need it.

`ShouldBindWith` does not check for saved body bytes,I can only use `ShouldBindBodyWith` after using `ShouldBindBodyWith` once. `ShouldBindJSON`，`ShouldBindXML` and `ShouldBindYAML` complain EOF

PR #1695, opened 2018-12-12
ci: exit 1 when build fail

thinkerou:


thinkerou:
@appleboy please help review, thanks!

PR #1677, opened 2018-12-01
Add LoggerWithFormatter method

sairoutine:
# Problems
I want to define gin's log format.
Currently, instead of using the default Logger, I can define the log format by defining my own Logger function.

However, my own Logger function must include redundant codes similar to the default Logger.

ref. https://github.com/gin-gonic/gin/issues/1581
# Proposal
This PR solves the problem by adding new methods, `LoggerWithFormatter` and `LoggerWithConfig`.

example code
```
r := gin.New()
r.Use(Recovery())
r.Use(LoggerWithFormatter(func(param gin.LogFormatterParams) string {
	// my own format
	return fmt.Sprintf("[GIN] %v | %3d | %13v | %15s | %-7s  %s | %s\n%s",
		param.TimeStamp.Format("2006/01/02 - 15:04:05"),
		param.StatusCode,
		param.Latency,
		param.ClientIP,
		param.Method,
		param.Path,
		SomeMyOwnParam(param.Request), // username or payload information
		param.ErrorMessage,
	)
}))
```

refs.
gorilla's LogFormatter
http://www.gorillatoolkit.org/pkg/handlers#LogFormatter

echo's LoggerWithConfig
https://echo.labstack.com/middleware/logger
# At end
If you prefers to this proposal, I will add unit tests and documents.

Thanks.

PR #1706, opened 2018-12-20
Plural is "Paths", not "Pathes"

sairoutine:
In https://github.com/gin-gonic/gin/pull/1677, I  misspelled LoggerConfig's SkipPaths.

**Note**
This change breaks backward compability, but adding LoggerConfig is not released yet. This change affects only the user which is using master branch.


sairoutine:
Thanks!

Issue #1678, opened 2018-12-03
Panic runtime error: index out of range in context.go:108


Collapsik:
After commit b97ccf3a43d2901d295870b5876277bf478a4909 I catch panic that never been before. In my case length of c.handlers equal 1 and c.index equal 1.

![image_2018-12-03_11-55-57](https://user-images.githubusercontent.com/16224266/49366578-4ed68400-f6f2-11e8-9ddf-96a4723c1daf.png)

thinkerou:
@Collapsik can you post your test code? I use it reproduce the issue, thanks!

Kirides:
I get the same / a simmilar issue. It appears to happen _after_ my handler returned

Code: https://pastebin.com/kM6UwyB3

thinkerou:
Sorry, I still can't reproduce it.

vkd:
I have reproduced it.

It happens because during [`func (c *Context) Next() {`](https://github.com/gin-gonic/gin/blob/master/context.go#L106) method it iterates over c.handlers with precalculated size of `c.handlers`.
```
func (c *Context) Next() {
	c.index++
	for s := int8(len(c.handlers)); c.index < s; c.index++ {
		c.handlers[c.index](c)
	}
}
```
But [`func (group *RouterGroup) StaticFS(...) ... {`](https://github.com/gin-gonic/gin/blob/master/routergroup.go#L172) rewrite `c.handlers` to slice with other size than we calculated in `c.Next(...)` (https://github.com/gin-gonic/gin/blob/master/routergroup.go#L198)
```
func (group *RouterGroup) createStaticHandler(relativePath string, fs http.FileSystem) HandlerFunc {
	...
	return func(c *Context) {
		...
		// Check if file exists and/or if we have permission to access it
		if _, err := fs.Open(file); err != nil {
			c.Writer.WriteHeader(http.StatusNotFound)
			c.handlers = group.engine.allNoRoute
			// Reset index
			c.index = -1
			return
		}

		fileServer.ServeHTTP(c.Writer, c.Request)
	}
}
```

thinkerou:
#1744

PR #1716, opened 2018-12-26
fix StaticFS() get "index out of range" if file not found (#1678)

vkd:
Fixing #1678

We cannot rewrite `c.handlers` in handler during iterate over it on `c.Next(...)`


thinkerou:
revert #1663 
cc @MetalBreaker

MetalBreaker:
I don't use Gin anymore. Do whatever you have to.

Issue #1590, opened 2018-10-13
How do I force logs to have color when writing files?


hel2o:
Now, when my log is written to the file, there is no color.

thinkerou:
why? now only output color log on term.

Issue #1547, opened 2018-09-14
not print color log when go version is 1.10/1.11


thinkerou:
- With issues:
  - Use the search tool before opening a new issue.
  - Please provide source code and commit sha if you found a bug.
  - Review existing issues and provide feedback or react to them.

- go version: go 1.11
- gin version (or commit ref): master
- operating system: mac

## Description
Here can see go 1.8 1.9 1.10 1.11 print log: https://travis-ci.org/gin-gonic/gin/builds/428455260?utm_source=github_status&utm_medium=notification

when go version is 1.9 it will print log:

![image](https://user-images.githubusercontent.com/4488261/45535407-f4cfcc00-b830-11e8-96a3-b5d516f3e8c2.png)

else print log:

![image](https://user-images.githubusercontent.com/4488261/45535437-0f09aa00-b831-11e8-9d7c-0acd99d7fd41.png)

thinkerou:
`println(isatty.IsTerminal(os.Stdout.Fd()))` on https://github.com/gin-gonic/gin/blob/master/logger.go#L60 will print `false` on go 1.11, but `println(isatty.IsTerminal(os.Stdout.Fd()))` on single file will print `true` on go 1.11. I don't know why?

thinkerou:
ping @mattn @thesyncim thanks

mattn:
I didn't change anything for mac. (as far as I remember)

thinkerou:
@mattn thank you for yourreply.

BrandonZacharie:
I got colors on macOS High Sierra and now Mojave. I only lose colors when running [gin](https://github.com/codegangsta/gin) (i.e. `gin -p 5000 -i run main.go`) which wraps the process.

kaustubhmallik:
@thinkerou will you be able to pick it, otherwise, I would love to help.

thinkerou:
@kamronbatman welcome

PR #1724, opened 2018-12-29
Add a function to force color in console output

orobardet:
Add a function `ForceConsoleColor`, like `DisableConsoleColor` but to force coloring the output.

It usefull when some IDE's integrated console (like IntelliJ or Goland) are not detected as TTY, but can display colors.

Also helps if one want to output color in log file (#1590) and as a workaround for #1547.

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,9 +215,6 @@ $ go build -tags=jsoniter .
 
 ```go
 func main() {
-	// Disable Console Color
-	// gin.DisableConsoleColor()
-
 	// Creates a gin router with default middleware:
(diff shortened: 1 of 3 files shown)


thinkerou:
@orobardet how to use it? and what distinguish it from `disableColor`?

orobardet:
> @orobardet how to use it? and what distinguish it from `disableColor`?

I added a new section in the README about controlling the colorization of log outputs, including the existing `DisableConsoleColor` and the new `ForceConsoleColor`.

PR #1729, opened 2019-01-02
Add context.HandlerNames()

lggomez:
This PR adds a HandlerNames method which returns all registered handles for the context in descending order


thinkerou:
when use `HandlerNames`? thanks!

lggomez:
Hi @thinkerou! at work we have dozens of large services which include both ad-hoc middlewares and base ones which come from an internal framework that dynamicly injects metrics and infrastructure related middlewares upon app initialization

In some cases and during testing, problems may arise due to invalid or wrong configurations on containers that result in different middleware pipelines being applied to those instances. In one specific case, we had an unhandled panic on a goroutine that was bringing down the service, and this helped to discard configuration issues on both the app server and the framework configuration (we've had those in the past) before I moved on to review the affected service implementation

tl;dr: It's useful to dump all the active handlers when diagnosing production issues

PR #1733, opened 2019-01-05
allow ignoring field on form mapping

bazaglia:
As discussed in issue #610 when using `c.Bind()`, sometimes we don't want to bind certain fields from the user request. While this can be done adding `json:"-"` to field tag, the same can't be achieved for multipart/formdata requests.

This PR allows setting in the tag `form:"-"`, which solves the described issue.


bazaglia:
Tests are actually passing, but even on master branch TravisCI tests are failing because of a random freeze https://travis-ci.org/gin-gonic/gin/builds/475558492

thinkerou:
not `json:"-"` but `binding:"-"` which also can use for multipart/formdata, thanks!

bazaglia:
@thinkerou hey, thanks for your comment.

problem is, as my comment here https://github.com/gin-gonic/gin/issues/610#issuecomment-451626124, gin uses binding for validation. That means the field still decoded by `c.Bind` even when setting `binding:"-"`. That was I added `form:"-"`, that's corresponding to `json:"-"`, but for forms.

In your example, when printing `form. Password`, it won't return a nil pointer (`<nil>`), but a memory address (like `0xc00001e510` that points to an empty string). This is really anoying specifically when saving data to a SQL database (it will save a empty string, not null)

binding/binding_test.go
```diff
@@ -278,6 +282,12 @@ func TestBindingFormForTime2(t *testing.T) {
 		"", "")
 }
 
+func TestFormBindingIgnoreField(t *testing.T) {
+	testFormBindingIgnoreField(t, "POST",
+		"/", "/",
+		"foo=bar", "")
```
vkd: Your test is not fail without your changes in `binding/form_mapping.go`.
```suggestion
		"-=bar", "")
```

bazaglia: That's right. I just force-pushed with this change. I run tests and they are still passing. 👍

hotcoffie:
its not working properly now. When I added form:"-" to my field, the field is still being assigned a value regardless of whether I use Bind or ShouldBind.

```go
type Astruct struct {
	Aa string  `form:"-" json:"aa"`
	Bb *string `form:"-" json:"bb"`
}

func Init(r *gin.Engine) {

	api := r.Group("/api")
	{
		api.GET("/ping", func(c *gin.Context) {
			c.JSON(http.StatusOK, "pong")
		})

		api.POST("/test", func(ctx *gin.Context) {
			var a Astruct
			ctx.Bind(&a)
			fmt.Println(a.Aa)
			fmt.Println(*a.Bb)
			ctx.AbortWithStatus(http.StatusOK)
		})
	}
}
```
```
POST http://127.0.0.1:9527/api/test
Content-Type: application/json

{
	"aa": "a",
	"bb": "b"
}
```
```
a
b
2023-09-05T01:30:16.577+0800	DEBUG	utils/logger.go:43	gin	{"method": "POST", "path": "/api/test", "status": 200, "ip": "127.0.0.1", "elapsed": "243.845µs"}
```

PR #1752, opened 2019-01-27
Add response size to LogFormatterParams

phacops:
Here is a change to make the response body size an available log formatter param.


sairoutine:
It looks good to me. 
To add the unit test will make your PR greater.

PR #1779, opened 2019-02-23
make context.Keys available as LogFormatterParams

MusicAdam:
This makes the context available in LogFormatterParams and therefore allows access to context variables as part of the log output formatting.


thinkerou:
@MusicAdam why need it? can you post one usage scenario? thanks!

thinkerou:
closing, thanks!

MusicAdam:
@thinkerou Its useful for logging context variables, such as a request-id. It makes the logger more robust.

sairoutine:
Instead of LogFormatterParams containing the context, it is better to add the its parameters or value to LogFormatterParams's field directly.

The context in Logger is useful. However, unnecessary methods of context can be used in the Logger. (For example `c.Abort()`, `c.JSON()`, `c.Next()` etc...)

Calling these functions may cause the unexpected bugs.

MusicAdam:
> Instead of LogFormatterParams containing the context, it is better to add the its parameters or value to LogFormatterParams's field directly.
> 
> The context in Logger is useful. However, unnecessary methods of context can be used in the Logger. (For example `c.Abort()`, `c.JSON()`, `c.Next()` etc...)
> 
> Calling these functions may cause the unexpected bugs.

This is a good point. I have updated so that context.Keys are passed instead. Since the Keys field is exported, I think it is safe to pass it directly instead of something more complicated like a copy or a separate accessor struct to allow Key retrieval.

sairoutine:
Great!

PR #1797, opened 2019-03-05
add support arrays on mapping

vkd:
```
var s struct {
	Nums [2]int `form:"nums"`
}
req := formPostRequest("", "nums=3&nums=8")
Form.Bind(req, &s)
```


binding/binding_test.go
```diff
@@ -1430,3 +1430,26 @@ func TestBindingTimeDuration(t *testing.T) {
...
+func TestBindingArray(t *testing.T) {
+	var s struct {
+		Nums [2]int `form:"nums,default=4"`
+	}
+
+	// default
+	req := formPostRequest("", "")
+	err := Form.Bind(req, &s)
+	assert.NoError(t, err)
+	assert.Equal(t, [2]int{4, 0}, s.Nums)
```
thinkerou: It's `[2]int{4, 0}`? It's not `[2]int{4, 4}`?

vkd: Yes, exactly. Because `default` tag gives only the one value and the second is just default value by `int` type.

Like in this case (TestBindingSliceDefault) : https://github.com/gin-gonic/gin/pull/1797/files/04507994072f9844b1d631ca15774a0d1fb80ac3#diff-ac17484cce1c0674a6dc181af5ef1edaL1383

thinkerou: OK, thanks. Maybe `[2]int{4, 4}` is better.

vkd: Could you explain why you think so?

vkd: In my mind better solution is to find a way to fill default by multivalues. Like `default=4,6`.

thinkerou: If I set default value, I should want to set all value is default value. If only first set default value, what's the meaning? It's kind of weird.

vkd: What do you think if I return error from mapping like `["4"] is not valid value for [2]int`?

thinkerou: About the below:
```
Nums [2]int `form:"nums,default=4"`
```
I think it should set default value, like this: `[2]int{4, 4}`.

And the below:
```
Nums [3]int `form:"nums,default=[4,2]"`
```
it should set default value, like this: `[3]int{4, 2, 0}`.

do you think?

Not recommend return error.

vkd: If add the support of working with array/slice in `default` tag it will give a little overhead and increase difficult of code. I think it is not a good value for that excess functionality.

The type `[2]int` shows that the variable has exactly 2 integers (like `type point struct{X, Y int}`). And in the case when we have different len of integers, indeed, we have another type that cannot match with the previous one. I prefer return an error because if `default=4` sets an array to `[2]int{4, 4}` it will be for me like unexpected behavior.

In summary, now the easiest way is just return an error if length of values is not equal with length of array. It allows to us improve it later without break compatibility.

thinkerou: got it.

PR #1802, opened 2019-03-08
IsTerm flag should not be affected by DisableConsoleColor method.

sairoutine:
It is strange behavior that `DisableConsoleColor` turns the `IsTerm` flag false.

This PR contains the following changes.
 - `IsTerm` represents whether is the output destination console.
 - Default logger distinguishes color by using new method `IsOutputColor`.
 - If `ForceConsoleColor` and `DisableConsoleColor` are called together, Later one is valid.

**Note**
This change breaks backward compability, but `LogFormatterParams` and `ForceConsoleColor` is not released yet. This change affects only the user which is using master branch.


logger.go
```diff
@@ -14,17 +14,24 @@ import (
...
-	forceColor   = false
+	green            = string([]byte{27, 91, 57, 55, 59, 52, 50, 109})
+	white            = string([]byte{27, 91, 57, 48, 59, 52, 55, 109})
+	yellow           = string([]byte{27, 91, 57, 48, 59, 52, 51, 109})
+	red              = string([]byte{27, 91, 57, 55, 59, 52, 49, 109})
+	blue             = string([]byte{27, 91, 57, 55, 59, 52, 52, 109})
+	magenta          = string([]byte{27, 91, 57, 55, 59, 52, 53, 109})
+	cyan             = string([]byte{27, 91, 57, 55, 59, 52, 54, 109})
+	reset            = string([]byte{27, 91, 48, 109})
+	consoleColorMode = autoColor
```
sairoutine: What happens when `ForceConsoleColor` is called after `DisableConsoleColor`?
This change makes the behavior easy understanding and the source code readable.

thinkerou:
About color I think that have only two case:
1. Not use color, even` isTerm=true`.
2. Use color, need `isTerm=true`.
so we only need one switch to indicate whether use color.
Default use color when isTerm=true, or not use.

sairoutine:
@thinkerou 
Your suggested usecases are true. Therefore, the following usecases are seldom occured.

> - If `ForceConsoleColor` and `DisableConsoleColor` are called together, Later one is valid.

This PR's purposes are two things.
 - It is strange behavior that `DisableConsoleColor` turns the `IsTerm` flag false.
 - `disableColor` and `forceColor` are redundant variables. To consolidate into `consoleColorMode` makes gin's code more readable.

thinkerou:
@sairoutine I hope simple.
gin default use color on Term mode.
If user calls `gin.DisableConsoleColor()`, not use color.

sairoutine:
@thinkerou 
>gin default use color on Term mode.
>If user calls gin.DisableConsoleColor(), not use color.

This PR does not change this behavior.

If the user does not call `gin.DisableConsoleColor()` or `gin.ForceConsoleColor()`, gin decides to use color by Term mode.

If the user calls `gin.DisableConsoleColor()`, gin does not use color whatever `isTerm=true` or `isTerm=false`.

thinkerou:
@sairoutine 
>This PR does not change this behavior.

Yes, I know. So I hope change it release gin1.4 before. do you think? thanks!

sairoutine:
@thinkerou 
Thanks! I also think this PR should be merged before gin 1.4 release.

thinkerou:
@sairoutine I hope the pull request change to:
> gin default use color on Term mode.
> If user calls gin.DisableConsoleColor(), not use color.
can you recommit one patch? thanks!

sairoutine:
@thinkerou 
It seems that this PR is according to this. Could you tell me what should I do?

>gin default use color on Term mode.
>If user calls gin.DisableConsoleColor(), not use color.
>can you recommit one patch? thanks!

thinkerou:
@sairoutine OK, I see #1724 content about `forceConsoleColor`, you not need to do what. Thanks!

sairoutine:
Thanks! Yes, This PR is related to #1724.

logger.go
```diff
@@ -115,10 +122,15 @@ func (p *LogFormatterParams) ResetColor() string {
 	return reset
 }
 
+// IsOutputColor indicates whether can colors be outputted to the log.
+func (p *LogFormatterParams) IsOutputColor() bool {
+	return consoleColorMode == forceColor || (consoleColorMode == autoColor && p.IsTerm)
```
thinkerou: `IsTerm bool` need to export?

sairoutine: I suppose so. Done. https://github.com/gin-gonic/gin/pull/1802/commits/d6440ae2499376b7adaa3ae24ccdb10faaa241a2

thinkerou: thanks!

PR #1803, opened 2019-03-08
Propagate request context behaviour to gin context methods implementing stdlib context.Context interface.

xakep666:
It makes possible to correctly cancel handlers when request cancelled (connection closed, etc.)
Try to extract value from request context when it was not found in gin context


thinkerou:
@xakep666 thank you commit, please see #1736

PR #1817, opened 2019-03-18
Clean the Request Path early

### LINKED ISSUE #1644: Multiple slashes in path is not working
## Description

Multiple slashes in path can't be routed!

In stdlib http, the [ServeMux](https://golang.org/src/net/http/server.go?s=40985:41006#L2313) use `cleanPath` on `path`. 

image

## Screenshots

image

dmarkham:
This will reduce the number of times we have to do a redirect
and allow multiple slashes in path to be routed!
fixes #1644


dmarkham:
Looks like the 1.6.x tests are failing. Are we planning on keeping 1.6 supported? 

If so I will need to rework `performRequest`  a little bit.   `httptest.NewRequest` is built for testing the server side requests and better supports the testing environment. 
For example `http.NewRequest`  given a url of `//foo/bar` will interpret `foo` is the host in a url *not* part of the path. `httptest.NewRequest` knows exactly what to do and is the documented way since version 1.7.x.     Thoughts and  Suggestions are very welcome.

thinkerou:
@dmarkham 
> Are we planning on keeping 1.6 supported?

Hi, we will remove go1.6/1.7 supported at gin1.4 or gin1.5.

thinkerou:
waiting for #1631

PR #1829, opened 2019-03-26
refactor(form_mapping.go): mapping multipart request

vkd:
Refactor mapping to interface `setter` - for supporting a multipart request.

Benchmarks of a simple form mapping (with many fields and one field respectively):
```
⇒  benchcmp old.bch new.bch
benchmark                  old ns/op     new ns/op     delta
BenchmarkMapFormFull-4     2332          2358          +1.11%
BenchmarkMapFormName-4     218           235           +7.80%

benchmark                  old allocs     new allocs     delta
BenchmarkMapFormFull-4     12             12             +0.00%
BenchmarkMapFormName-4     1              1              +0.00%

benchmark                  old bytes     new bytes     delta
BenchmarkMapFormFull-4     200           200           +0.00%
BenchmarkMapFormName-4     8             8             +0.00%
```

### DIFF
--- a/binding/form.go
+++ b/binding/form.go
@@ -72,7 +72,8 @@ var (
 	multipartFileHeaderStructType = reflect.TypeOf(multipart.FileHeader{})
 )
 
-func (r *multipartRequest) Set(value reflect.Value, field reflect.StructField, key string, opt setOptions) (isSetted bool, err error) {
+// TrySet - try to set a value by the multipart request with the binding a form file
+func (r *multipartRequest) TrySet(value reflect.Value, field reflect.StructField, key string, opt setOptions) (isSetted bool, err error) {
 	if value.Type() == multipartFileHeaderStructType {
 		_, file, err := (*http.Request)(r).FormFile(key)
 		if err != nil {
--- a/binding/form.go
+++ b/binding/form.go
@@ -66,6 +66,8 @@ func (formMultipartBinding) Bind(req *http.Request, obj interface{}) error {
 
 type multipartRequest http.Request
 
+var _ setter = (*multipartRequest)(nil)
+
 var (
 	multipartFileHeaderStructType = reflect.TypeOf(multipart.FileHeader{})
 )
--- a/binding/form.go
+++ b/binding/form.go
@@ -4,7 +4,11 @@
 
 package binding
 
-import "net/http"
+import (
+	"mime/multipart"
+	"net/http"
+	"reflect"
+)
 
 const defaultMemory = 32 * 1024 * 1024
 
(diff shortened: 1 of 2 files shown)


binding/form.go
```diff
@@ -53,13 +57,30 @@ func (formMultipartBinding) Bind(req *http.Request, obj interface{}) error {
...
+	return validate(obj)
+}
+
+type multipartRequest http.Request
+
+var (
+	multipartFileHeaderStructType = reflect.TypeOf(multipart.FileHeader{})
+)
+
+func (r *multipartRequest) Set(value reflect.Value, field reflect.StructField, key string, opt setOptions) (isSetted bool, err error) {
```
thinkerou: check: `var _ setter = &multipartRequest`

binding/form_mapping.go
```diff
@@ -57,11 +28,25 @@ func mapForm(ptr interface{}, form map[string][]string) error {
...
+	return mappingByPtr(ptr, formSource(form), tag)
+}
+
+type setter interface {
+	Set(value reflect.Value, field reflect.StructField, key string, opt setOptions) (isSetted bool, err error)
+}
+
+type formSource map[string][]string
+
+func (form formSource) Set(value reflect.Value, field reflect.StructField, tagValue string, opt setOptions) (isSetted bool, err error) {
```
thinkerou: check: `var _ setter = &formSource`
and add node for export method/class/interface.

binding/form_mapping.go
```diff
@@ -57,11 +28,25 @@ func mapForm(ptr interface{}, form map[string][]string) error {
 var emptyField = reflect.StructField{}
 
 func mapFormByTag(ptr interface{}, form map[string][]string, tag string) error {
-	_, err := mapping(reflect.ValueOf(ptr), emptyField, form, tag)
+	return mappingByPtr(ptr, formSource(form), tag)
+}
+
+type setter interface {
```
thinkerou: add some nodes?

binding/form.go
vkd: What do you mean when you talk about "node"?
Btw, this method is not exported because the type `multipartRequest` is not exported, and I see no warnings about this from any linters.

thinkerou: @vkd sorry, word typo, `note`.
Yes, it's not exported. We should add some usage note for new and important interface/method/class even if un-exported. thanks!

vkd: Ok, thanks. I've added a couple of comments.

binding/form.go
```diff
@@ -72,7 +72,8 @@ var (
 	multipartFileHeaderStructType = reflect.TypeOf(multipart.FileHeader{})
 )
 
-func (r *multipartRequest) Set(value reflect.Value, field reflect.StructField, key string, opt setOptions) (isSetted bool, err error) {
+// TrySet - try to set a value by the multipart request with the binding a form file
```
thinkerou: please not use `TrySet - ` style, thanks! like: `TrySet tries to set ...`

binding/form_mapping.go
thinkerou: please not use `// setter -` style, thanks!

binding/form_mapping.go
```diff
@@ -33,14 +33,15 @@ func mapFormByTag(ptr interface{}, form map[string][]string, tag string) error {
...
+	TrySet(value reflect.Value, field reflect.StructField, key string, opt setOptions) (isSetted bool, err error)
 }
 
 type formSource map[string][]string
 
 var _ setter = formSource(nil)
 
-func (form formSource) Set(value reflect.Value, field reflect.StructField, tagValue string, opt setOptions) (isSetted bool, err error) {
+// TrySet - try to set a value by request's form source (like map[string][]string)
+func (form formSource) TrySet(value reflect.Value, field reflect.StructField, tagValue string, opt setOptions) (isSetted bool, err error) {
```
thinkerou: the same above, thanks!

PR #1842, opened 2019-04-03
binding: move tests of mapping to separate test file

vkd:
Made a 100% coverage of `form_mapping.go` from the `form_mapping_test.go` file.
And rewrote to the more readable test functions.


dmarkham:
@vkd  are you able to tell whats up with that one failing test?

PR #1843, opened 2019-04-04
Remove go1.6, go1.7, go1.8 support

dmarkham:
I'm aware removing `go1.6`/`go1.7` was scheduled for `v1.5.0` but given the amount of time that has passed we now have `go1.12` so supporting only `go1.9+` is older than `go` supports.


thinkerou:
please see #1631 thanks!

PR #1849, opened 2019-04-06
v1.4.0

dmarkham:
This is a starting point for the v1.4.0  release issue #1732.

Looking for feedback or patches for anything that needs to be changed.


dmarkham:
So after taking a stab at getting the `govender` to match `go modules` I'm not yet able to to get `go1.6` to work.  The main reason is we have moved from version`v0.04` ->  `v0.0.6` of `github.com/mattn/go-isatty` This package is now using the  `golang.org/x/sys/unix` pacakage  that then  uses `runtime.KeepAlive` this is *not* in `go1.6`.  So This leaves us a few options. 

1. downgrade `github.com/mattn/go-isatty`
2. drop support for `go1.6` and `go1.7` in  gin `1.4.0`
3. use different versions of `github.com/mattn/go-isatty` depending on your choice of dependency manager/go version. 

My preference is to drop `go1.6` and `go1.7` in `1.4.0`.

Looking for some direction on how to move forward.

Issue #1847, opened 2019-04-06
Huge dependency


j0hnwhyte:
Using go 1.12, latest version of gin to date, I analysed my executable and I see that 11MB are occupied by the dependency `github.com/ugorji/go/codec`. Does really gin need all of this package?

tqcenglish:
some result through [goweight](https://github.com/jondot/goweight)

sairoutine:
`github.com/ugorji/go/codec` is used to output msgpack.
It is huge because `codec` supports not only msgpack but also cbor, json.

It should be better we thinking to use other msgpack libraries instead of it.

Hades32:
Just realized just the same thing. That's ridiculously large for a hello world http server.

I would argue that MsgPack support shouldn't be part of core Gin anyway. It's not a super common format like Json. Some goes probably for `github.com/golang/protobuf` ~2MB...

IMHO it would make more sense to have a gin-msgpack and gin-protobuf dependency

pollen5:
i agree, those dependencies should be someway optional or as another plugin package, in most cases i don't think a user would need those dependencies.

j0hnwhyte:
Good frameworks are built up in modules. As we don't expect gin to have other useful stuff like jwt, it shouldn't include by default a relatively more obscure encoding for no reason, in my own opinion.

dmarkham:
So `go` in general is just a little over the top in executable size. If we were all prioritizing saving 2MB we would be writing in rust or c.

I don't really like the build tags option either. Last thing I want to think about is all the build tags we end up adding add then having to support in the future.

Now that binding has been refactored into a nice interface. Seems like we should be able to get all none builtin encoding's moved out into something that registers itself with `gin` on a per app basis maybe similar to how `sql` drivers do it. We can do better!

sapk:
@dmarkham the build flag is mainly to have non-breaking optimization but a system like sql drivers should be better and should not be that difficult to setup. The most difficult point would be the current select of content-type in the binding module. The rest could already be 'pluggable' but that would break reverse compat with older version and would need a v2 versioning.

dmarkham:
@sapk technically I don't think msgpack / protobuff have been in a release yet. They are being proposed  to be included in `v1.4.0` and that is not out yet. Maybe they should be the first to use a new ''pluggable" version, vs the first to to start a trend on being excluding via tags?

guonaihong:
If you remove github.com/ugorji/go/codec, it will automatically build faster for me.

[...]

PR #1856, opened 2019-04-17
Modular deps for binding and render

sapk:
Following the similar pattern as sql drivers.
I extract xml, json, msgpack, protobuf and yaml into separate module.
This PR may break compat with previous versions.

For example: https://github.com/sapk-fork/gin-examples/blob/test-modular/basic/main.go

I think we can maybe re-add json and xml as default.

Related #1847 & #1852


sapk:
For details:
- Context function calls are unchanged
- If the need module is not loaded it will panic for method like `(c *Context) MsgPack` or `(c *Context) BindYAML(obj interface{})`

dmarkham:
@sapk  :clap: 

> For example: https://github.com/sapk-fork/gin-examples/blob/test-modular/basic/main.go
I kinda like the way this feels only opting into what you want really nice :+1: 

> 
> I think we can maybe re-add json and xml as default.

I do like the idea of the "builtins" being there by default. But would that allow the user to replace the xml/json with a different (faster) binding/render for json/xml if we are including it by default?

I'm really interested in the core teams thoughts on this idea is!
 
:+1: on the user's interface. 
I'm still looking at the code to get a better feel for how you pulled this off.

sapk:
@dmarkham Yes you can load a module that fill the same id in the respective common List to be call in place of the gin one.
If you want to already use a custom binding or render this is already possible with context method `MustBindWith/ShouldBindWith` and `Render` that just need a object that provide the respective interfaces.

zzjin:
any update discuss?

PR #1881, opened 2019-05-05
add Forward support

fanchenggang:
Engine is private .
But sometimes we need to call HandleContext inside a function .
like this:
```go
func HelloWord(c *gin.Context){
	c.Request.URL.Path = "/test2"
	c.Forward()
}
```


dmarkham:
@fanchenggang I think I understand what your are tying to do.

This being exposed is *interesting*, even at the server level seems wacky, In most cases you can just point to the same function in different routes to get this "soft" redirect feature without all the URL path hacking and redirect loop issues. After catching up on #709 and the reason we have  `s.HandleContext()` I do better understand how it's useful.

Doing this down at the context level as you have exposed, may have more gotchas to look out for than 
 a infinite path loop. Anything you have already written to the `writer` will be wiped out, anything your `handlerFunc` or your  middleware did to the `c.Request.Body` would have consumed it and it will not be there for your `c.Forward()` handler's and middleware. 

I wish we had a easy way to build this feature without creating a traps for the people that really should have just do a `Redirect` in the first place. 

@fanchenggang  do you mind sharing the use case your solving with this change.

fanchenggang:
@dmarkham  Maybe I didn't think it through, I want to forward rather than redirect

dmarkham:
> @dmarkham Maybe I didn't think it through, I want to forward rather than redirect

Understood,  Your idea is better than anything i have come up with, it just has some edge cases we need to better understand.

PR #1893, opened 2019-05-10
Now you can parse the inline lowercase start structure

guonaihong:
``` golang
package main

import (
	"github.com/gin-gonic/gin"
)

type appkey struct {
	Appkey string `json:"appkey" form:"appkey"`
}

type Query struct {
	Page int `json:"page" form:"page"`
	Size int `json:"size" form:"size"`
	appkey
}

func main() {

	router := gin.Default()
	router.POST("/login", func(c *gin.Context) {

		var q2 Query

		if c.ShouldBindQuery(&q2) == nil {
			c.JSON(200, &q2)
		}
	})
	router.Run(":8088")
}
```
http client:

old:
``` bash
curl -X POST "127.0.0.1:8088/login?appkey=china&page=1&size=10"
{"page":1,"size":10,"appkey":""}
```
now:
``` bash
curl -X POST "127.0.0.1:8088/login?appkey=china&page=1&size=10"
{"page":1,"size":10,"appkey":"china"}
```

### DIFF
--- a/binding/form_mapping.go
+++ b/binding/form_mapping.go
@@ -70,12 +70,14 @@ func mapping(value reflect.Value, field reflect.StructField, setter setter, tag 
 		return isSetted, nil
 	}
 
-	ok, err := tryToSetValue(value, field, setter, tag)
-	if err != nil {
-		return false, err
-	}
-	if ok {
-		return true, nil
+	if !(vKind == reflect.Struct && field.Anonymous) {
+		ok, err := tryToSetValue(value, field, setter, tag)
+		if err != nil {
+			return false, err
+		}
+		if ok {
+			return true, nil
+		}
 	}
 
 	if vKind == reflect.Struct {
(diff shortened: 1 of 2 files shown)


binding/form_mapping.go
```diff
@@ -70,20 +70,23 @@ func mapping(value reflect.Value, field reflect.StructField, setter setter, tag
 		return isSetted, nil
 	}
 
-	ok, err := tryToSetValue(value, field, setter, tag)
-	if err != nil {
-		return false, err
-	}
-	if ok {
-		return true, nil
+	if !(vKind == reflect.Struct && field.Anonymous) {
```
thinkerou: thanks! I suggest change it to 
`if vKind != reflect.Struct || !field.Anonymous {`

guonaihong: Thank you, I have modified it.

PR #1902, opened 2019-05-16
feat(gin,run): add expose out http.Server can use Server config

LinuxForYQH:
I want  make gin more easy and comfortable to use http.Server configuration
Before
```go
r := gin.Default()
    c.JSON(200, gin.H{
      "message": "pong",})
})
server := &http.Server{Addr: ":8080",Handler:r}
server .ReadTimeout = time.Duration(time.Second*5)
server.ListenAndServe()
```
After comfortable and friendly
```go
r := gin.Default()
    c.JSON(200, gin.H{
      "message": "pong",})
})
server := &http.Server{ReadTimeout:time.Duration(time.Second*5),}
r.RunServer(server,"8080")
```

### DIFF
--- a/gin.go
+++ b/gin.go
@@ -295,6 +295,23 @@ func (engine *Engine) Run(addr ...string) (err error) {
 	return
 }
 
+// Run attaches the router to a http.Server and starts listening and serving HTTP requests.
+// You can use your http.server config like ReadTimeout ...
+// It is a shortcut for http.ListenAndServe(addr, router)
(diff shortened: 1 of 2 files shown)


dmarkham:
@LinuxForYQH  Thank you very much.

But I do not think adding a new method for this is so much better it's worth having a new methods to support it. If we do this I think we will also need a `TLS` version also.  If it were up to me I think this is a cool idea just too small of an improvement. :-1:

PR #1933, opened 2019-06-06
Drop Support for go1.8 and go1.9

dmarkham:
Here are the updates for fix #1912

PR #1939, opened 2019-06-10
Gin1.5 bytes.Buffer to strings.Builder

guonaihong:
Go1.10's strings.Builder structure saves memory
This is its implementation principle
```go
// String returns the accumulated string.
func (b *Builder) String() string {
    return *(*string)(unsafe.Pointer(&b.buf))
}

```

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -6,9 +6,10 @@ package gin
 
 import (
 	"fmt"
-	"github.com/gin-gonic/gin/internal/json"
 	"reflect"
 	"strings"
+
+	"github.com/gin-gonic/gin/internal/json"
 )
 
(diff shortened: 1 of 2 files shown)


thinkerou:
waiting #1933 merged.

guonaihong:
ok

errors.go
```diff
@@ -5,11 +5,10 @@
...
 
 import (
-	"bytes"
 	"fmt"
-	"reflect"
-
 	"github.com/gin-gonic/gin/internal/json"
+	"reflect"
+	"strings"
 )
```
thinkerou: please update import statement sort.

PR #1940, opened 2019-06-10
Modify status

vearne:
I want to develop a timeout middleware, but I' found only can do it by fix context.Status()


thinkerou:
please see #1606 thanks!

Issue #1731, opened 2019-01-03
[bug] panic: context: internal error: missing cancel error


maopao-691515082:
- go version: 1.11
- gin version (or commit ref):807368579f8939cedfa59ec689708754920f93e4
- operating system:linux

## Description
panic in package context:

panic: context: internal error: missing cancel error

goroutine 6 [running]:
context.(*cancelCtx).cancel(0xc000052180, 0xc000090000, 0x0, 0x0)
	/home/xiaotao.01/go-1.11/src/context/context.go:369 +0x258
context.(*timerCtx).cancel(0xc000052180, 0x0, 0x0, 0x0)
	/home/xiaotao.01/go-1.11/src/context/context.go:447 +0x4a
context.propagateCancel.func1(0xc00007e000, 0x207, 0x10000, 0x4de000, 0xc00000a080, 0x4dde40, 0xc000052180, 0xc0000700c0, 0xc000070120, 0x0)
	/home/xiaotao.01/go-1.11/src/context/context.go:281 +0x1f6
created by context.propagateCancel
	/home/xiaotao.01/go-1.11/src/context/context.go:271 +0x316
exit status 2

analysis:
gin use sync.pool to cache gin.Context objects:
```
func (engine *Engine) ServeHTTP(w http.ResponseWriter, req *http.Request) {
	c := engine.pool.Get().(*Context)
	c.writermem.reset(w)
	c.Request = req
	c.reset()

	engine.handleHTTPRequest(c)

	engine.pool.Put(c)
}
```

but a context objects should not be reused, if we use ginCtx as parent to make a new timeout-context(or deadline-context or cancel-context), the package context will start a goroutine to select Done() of parent and child context. if parent is done, cancel child by parent.Err(), but parent is a reusable ginCtx, if it's already using by another handler, the Err() is nil, then panic as above

in short, gin put ginCtx back to pool means "free this context", but it's still used by package context

The Context cannot be cached in the pool, otherwise the Context that is immediately fetched by another goroutine will be in an unknown state, which may lead to a Panic
    --by Vincen, my workmate

---

[...]

PR #2029, opened 2019-08-27
fix broken context.Context implementation

ekle:
The current implementation of `gin.Context` does satisfy the `context.Context` interface, but it returns only `default`/`nil` values.

This MR changes this to use the values from the underlying `http.Request` instead.

This allows for example to use `opencensus` middleware for cross service tracing which uses the context to transfer tracing-IDs.

### DIFF
--- a/context.go
+++ b/context.go
@@ -1038,14 +1038,14 @@ func (c *Context) SetAccepted(formats ...string) {
 // should be canceled. Deadline returns ok==false when no deadline is
 // set. Successive calls to Deadline return the same results.
 func (c *Context) Deadline() (deadline time.Time, ok bool) {
-	return
+	return c.Request.Context().Deadline()
 }
 
 // Done returns a channel that's closed when work done on behalf of this
 // context should be canceled. Done may return nil if this context can
 // never be canceled. Successive calls to Done return the same value.
 func (c *Context) Done() <-chan struct{} {
-	return nil
+	return c.Request.Context().Done()
 }
 
 // Err returns a non-nil error value after Done is closed,
@@ -1055,7 +1055,7 @@ func (c *Context) Done() <-chan struct{} {
 // Canceled if the context was canceled
 // or DeadlineExceeded if the context's deadline passed.
 func (c *Context) Err() error {
-	return nil
+	return c.Request.Context().Err()
 }
 
 // Value returns the value associated with this context for key, or nil
@@ -1066,8 +1066,9 @@ func (c *Context) Value(key interface{}) interface{} {
 		return c.Request
 	}
 	if keyAsString, ok := key.(string); ok {
-		val, _ := c.Get(keyAsString)
-		return val
+		if val, ok := c.Get(keyAsString); ok {
+			return val
+		}
 	}
-	return nil
+	return c.Request.Context().Value(key)
 }


segevfiner:
Wasn't this done before and was broken due to Gin caching the `gin.Context` objects?

ekle:
@segevfiner the current master contains just stubs for the context functions.
https://github.com/gin-gonic/gin/blob/6ece26c7c5ce36863599c9a897514e2a70d4021c/context.go#L1033-L1073
The `Value` function is just a mapping to some gin internal store, but not to the request context.

segevfiner:
Yeah I know. I remember looking into this myself and seeing that there was a reverted PR for it because it caused some bug due to `gin.Context` cache or something like that. https://github.com/gin-gonic/gin/pull/1736

ekle:
I looked at the https://github.com/gin-gonic/gin/pull/1736 
The revert was wrong. It worked perfectly fine.
The `panic` did just happen because they missed to call the `context.Copy` function.
Which is a very unusual requirement for `context.Context` but the gin documentation does say so.
https://github.com/gin-gonic/gin/blob/6ece26c7c5ce36863599c9a897514e2a70d4021c/context.go#L89-L91

bbiao:
@ekle How to avoid this bug? https://github.com/gin-gonic/gin/issues/1731

bbiao:
@ekle How to avoid this bug? https://github.com/gin-gonic/gin/issues/1731

ekle:
@bbiao it should not happen if `Copy()` is called, as stated in the documentation.
If it still would panic, either `Copy()` is not called or `Copy()` has a bug.

btw: not calling `Copy()` in such cases is a big security hole, as the pool would replace the underlying `http.Request` by reusing the `gin.Context` while still another instances would access it. The old instance would get access to a different new http.Request. So if it panics, it should, as this could be a big security problem. (the panic is only triggered in some cases)

bryanpaluch:
Running into this still, and its because the standard library creates a go routine when context.WithTimeout or context.WithDeadline is used. I can work around by spamming c.Copy() all over the place whenever I pass a *gin.Context to a function that may call context.WithTimeout but its a pretty lousy solution when all of my code will likely call a library that uses context.WithTimeout. 

Here's a breaking test.
`func TestConcurrentHandleContextCancel(t *testing.T) {
	router := New()
	router.GET("/example", func(c *Context) {
		ctx, cancel := context.WithTimeout(c, 1*time.Millisecond)
		defer cancel()
		<-ctx.Done()
		c.String(http.StatusOK, "it worked")
	})

	var wg sync.WaitGroup
	iterations := 200
	wg.Add(iterations)
	for i := 0; i < iterations; i++ {
		go func() {
			req, err := http.NewRequest("GET", "/example", nil)
			assert.NoError(t, err)
			ctx, cancel := context.WithCancel(req.Context())
			req = req.WithContext(ctx)
			w := httptest.NewRecorder()
			router.ServeHTTP(w, req)
			go func() {
				cancel()
			}()
			wg.Done()
		}()
	}
	wg.Wait()
}`
If you use c.Copy() on the context.WithTimeout line it won't panic. I can't imagine using c.Copy() all over the place is a good idea, we might as well remove Context pooling. I'm trying to work out middle ground patch.

Chi has had this issue as well,
https://github.com/go-chi/chi/pull/74
https://github.com/go-chi/chi/pull/75

Issue #2043, opened 2019-09-06
Context.Set(key, value) map not goroutine safe


legenove:
- With issues:
  - Use the search tool before opening a new issue.
  - Please provide source code and commit sha if you found a bug.
  - Review existing issues and provide feedback or react to them.

- go version:  v1.11.5
- gin version (or commit ref):
- operating system:

thinkerou:
#1391

PR #2046, opened 2019-09-06
context add RWMutex for key set and get Goroutine safe

legenove:
#2043 
add rwmutex to ensure the security when we use multiple goroutines.


thinkerou:
#1391

PR #2086, opened 2019-10-06
Relocate binding body tests

vkd:
Every test file should be related to a tested file.
Remove useless tests.


binding/json_test.go
```diff
@@ -0,0 +1,21 @@
...
+// Use of this source code is governed by a MIT style
+// license that can be found in the LICENSE file.
+
+package binding
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
thinkerou: need the package?

vkd: Fixed, I've added that package in vendor file.

That package helps us to show an useful test output on error cases. In case when we see assert's message of `err != nil`, the message about `field's value is not equals` says nothing helpful.

PR #2090, opened 2019-10-11
Remove unused Name() method from Binding interface

vkd:
The method Name() is not used in gin.


thinkerou:
I not suggest remove it

vkd:
@thinkerou Why do you think so?

PR #2093, opened 2019-10-12
Fix 'errcheck' linter warnings

vkd:


Dominik-K:
@vkd Which (meta-)linter are you using? It seems that it didn't find that the [error for `os.Remove()` isn't checked (in `RunUnix()`)](https://github.com/gin-gonic/gin/pull/2093/files#diff-703ef7478e12fbd8acd9d1af87bda71cR317), as well?

vkd:
@Dominik-K I use the [golangci-lint](https://github.com/golangci/golangci-lint)

PR #2094, opened 2019-10-12
Update go.sum file

vkd:


thinkerou:
@vkd what do you run command?

vkd:
@thinkerou `go test ./...` or `go mod tidy` commands make this change in go.sum file.

PR #2095, opened 2019-10-13
Add RecoveryWithWriterAndCallback

letsfire:
- With pull requests:

### DIFF
--- a/recovery.go
+++ b/recovery.go
@@ -33,6 +33,18 @@ func Recovery() HandlerFunc {
 
 // RecoveryWithWriter returns a middleware for a given writer that recovers from any panics and writes a 500 if there was one.
 func RecoveryWithWriter(out io.Writer) HandlerFunc {
+	return RecoveryWithWriterAndCallback(out, func(c *Context, err interface{}, brokenPipe bool) {
+		// If the connection is dead, we can't write a status to it.
+		if brokenPipe {
+			c.Error(err.(error)) // nolint: errcheck
+			c.Abort()
+		} else {
+			c.AbortWithStatus(http.StatusInternalServerError)
+		}
+	})
+}
+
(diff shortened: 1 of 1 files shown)


recovery.go
```diff
@@ -33,6 +33,18 @@ func Recovery() HandlerFunc {
...
+		if brokenPipe {
+			c.Error(err.(error)) // nolint: errcheck
+			c.Abort()
+		} else {
+			c.AbortWithStatus(http.StatusInternalServerError)
+		}
+	})
+}
+
+func RecoveryWithWriterAndCallback(out io.Writer, fn func(c *Context, err interface{}, brokenPipe bool)) HandlerFunc {
```
vkd: Please add a comment block for this function.

letsfire: I did it

vkd:
I don't like this semantic of the callback function.
Why err variable has type `interface{}`?
What is the `brokenPipe` variable for and what does it mean?
Especially without a detailed description.
Because this is a public function of gin framework, this means that everyone who uses it should be able to understand for what the function is and how to use it.

letsfire:
My english is broken, I can't express myself.   
I don't want the HTTP status code to be 500, i hope i can  customize it

```go
func (l *logger) GinRecovery() gin.HandlerFunc {
	return gin.RecoveryWithWriterAndCallback(
		l.getWriter(logrus.PanicLevel.String()),
		func(c *gin.Context, err interface{}, brokenPipe bool) {
			var res errs.Error
			switch err.(type) {
			case error:
				res.Use(err.(error))
			default:
				res = errs.NewErrorf(500000, "panic error, %v", err)
			}
			c.JSON(http.StatusOK, res)
		},
	)
}

// the response maybe like this
{
    "code": 500***,
    "message": "the panic error description"
}
```

PR #2096, opened 2019-10-13
Add: gin.StyleLogger, and the method functions and unit test codes of it.

kerbalwzy:
Add a new struct type use for output the custom log information in gin.HanlderFunc with gin log style.

Unit testing result:
image

Preview of Actual Use Effect:
image

Usage demo code:
```go
package main

import (
	"github.com/gin-gonic/gin"
	"log"
)

var GlobalGinStyleLogger *gin.StyleLogger

func init() {
	GlobalGinStyleLogger = gin.NewGinStyleLogger(nil, nil)
}

func ginStylLoggerDemo(c *gin.Context) {
	GlobalGinStyleLogger.Fprintln(c, 500, "<custom log information here>")
	c.Status(200)
}

func main() {
	router := gin.Default()
	router.GET("/example", ginStylLoggerDemo)
	err := router.Run(":8080")
	if nil != err {
		log.Fatal(err)
	}
}
```


thinkerou:
@kerbalwzy thanks a lot! please see #1677

PR #2097, opened 2019-10-14
binding: add ShouldBindQuery2 and BindQuery2

guonaihong:
### Why mention this pr
In data binding
http header uses the SoupdBindHeader function, the tag is the header
JSON uses the SoupdBindJSON function, the tag is json
uri uses the ShoulddBindUri function, the tag is uri
In design, keep a tag and a function one-to-one correspondence.
However, both ShouldBind and ShoulddBindQuery use form tags.

This pr just separates the query from the form, based on the considerations below.

* Behind the good aspects plus new features, do not worry about breaking the form data binding.
* Used in conjunction with github.com/guonaihong/gout http client, people can easily bind data with gin and then forward the data to the next service node with gout

### demo
```go
package main

import (
	"github.com/gin-gonic/gin"
        "github.com/guonaihong/gout"
)

type testQuery struct {
	Size int    `query:"size"`
	Page int    `query:"page"`
	Ak   string `query:"ak"`
}

func nextSever() {
        r := gin.Default()

	r.GET("/query", func(c *gin.Context) {
		q := testQuery{}
		err := c.ShouldBindQuery2(&q)
		if err != nil {
			return
		}
		c.JSON(200, q)
	})
        r.Run(":1234")
}

func main() {
        go nextSever()
	r := gin.Default()

	r.GET("/query", func(c *gin.Context) {
		q := testQuery{}
		err := c.ShouldBindQuery2(&q)
		if err != nil {
			return
		}
                // Send to the next service
                code := 0 // http code
                err := gout.GET("127.0.0.1:1234/query").SetQuery(q).Code(&code).Do()
                if err != nil || code != 200 { /* todo Need to handle errors here */}
		c.JSON(200, q)
	})

	r.Run()
}

// http client
// curl '127.0.0.1:8080/query?size=10&page=20&ak=test'
```

### DIFF
--- a/binding/binding.go
+++ b/binding/binding.go
@@ -72,6 +72,7 @@ var (
 	XML           = xmlBinding{}
 	Form          = formBinding{}
 	Query         = queryBinding{}
+	Query2        = queryBinding2{}
 	FormPost      = formPostBinding{}
 	FormMultipart = formMultipartBinding{}
 	ProtoBuf      = protobufBinding{}
--- a/binding/binding_test.go
+++ b/binding/binding_test.go
@@ -35,12 +35,12 @@ type QueryTest struct {
 }
 
 type FooStruct struct {
-	Foo string `msgpack:"foo" json:"foo" form:"foo" xml:"foo" binding:"required"`
+	Foo string `msgpack:"foo" json:"foo" form:"foo" xml:"foo" query:"foo" binding:"required"`
 }
 
 type FooBarStruct struct {
 	FooStruct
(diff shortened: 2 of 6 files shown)


thinkerou:
what's the difference with `ShouldBindQuery` and `BindQuery2`?

guonaihong:
Use different structure tags, one is form and one is query

thinkerou:
that's not a good name, and you should consider break-end, do we really need the feature?

guonaihong:
I will close this pr first. This pr is not necessary from the perspective of gin.

PR #2112, opened 2019-10-28
FileFromFS

nikandfor:
I was missing feature of Serving file from `http.FileSystem`.

Use cases:
* Serve file by alias without redirection and opening file by hands.
* Return help page on root of `/api/` handlers.
* Choose `FileSystem` dynamically.


nikandfor:
Hi! I'm just remembering about my PR)

thinkerou:
need @appleboy review the PR, thanks1

baurine:
Hi @nikandfor , can you help add more detail for this part:

```
	var fs http.FileSystem = // ...
	router.GET("/fs/file", func(c *gin.Context) {
		c.FileFromFS("fs/file.go", fs)
	})
```

What the exact sample content for `// ...`, how `fs/file.go` locate in fs?

I tried many cases,  they all don't work for me, my sample code:

```
	endpoint := r.Group("/diagnose")
	endpoint.StaticFS("/assets", http.Dir("ui/build"))
	var fs http.FileSystem = http.Dir("ui/build")
	endpoint.GET("/reports/detail", func(c *gin.Context) {
		// c.File("ui/build/index.html") // this line code can work well
		c.FileFromFS("ui/build/index.html", fs)
	})
```

For the first parameter of `c.FileFormFS()`, I  tried `ui/build/index.html`, `./index.html`, `build/index.html`, all I failed to get the correct response, but the `c.File("ui/build/index.html")` can success.

Thank  you!

baurine:
After deep into the source code, I found just we can't use index.html, it will be redirected, the source code in fs.go#serverFile() method.

```go
// name is '/'-separated, not filepath.Separator.
func serveFile(w ResponseWriter, r *Request, fs FileSystem, name string, redirect bool) {
	const indexPage = "/index.html"

	// redirect .../index.html to .../
	// can't use Redirect() because that would make the path absolute,
	// which would be a problem running under StripPrefix
	if strings.HasSuffix(r.URL.Path, indexPage) {
		localRedirect(w, r, "./")
		return
	}
```

So I change the index.html name to aaa.html and works by following  code:

```go
	endpoint := r.Group("/diagnose")
	endpoint.StaticFS("/assets", http.Dir("ui/build"))
	var fs http.FileSystem = http.Dir("ui/build")
	endpoint.GET("/reports/detail", func(c *gin.Context) {
		c.FileFromFS("aaa.html", fs)
	})
```

nikandfor:
@baurine It's really interesting finding.
Would you find solution at your project code?
How do you think if we should add check at `gin.Context.FileFromFS`? like
```go
const indexPage = "index.html"

filepath = strings.TrimSuffix(filepath, indexPage)
```

baurine:
hi @nikandfor , I just rename my  index.html to report.html to workaround it...

nikandfor:
@thinkerou how do you think, do we need to add some fix to avoid such situations in the future?

ddvk:
this was quite annoying, trying to serve an embedded asset called `index.html` resulting in too many redirects, until finding out what was causing the problem. 

Maybe add a comment or throw an error

lenghan4real:
> @thinkerou how do you think, do we need to add some fix to avoid such situations in the future?

I think so, rename `index.html` is quite annoying.

PR #2118, opened 2019-10-30
Close files opened in static file handler

### LINKED ISSUE #2117: Bug: Static file handler does not close files
The static file handler (`RouterGroup.createStaticHandler`) does not close files opened via `fs.Open`:

[Permalink here](https://github.com/gin-gonic/gin/blob/393a63f3b020df89d42695064443760c7d0a0dc8/routergroup.go#L196).

This can lead leave file handles open and may prevent other processes from opening/modifying the file.

Shamus03:
Fixes #2117

### DIFF
--- a/routergroup.go
+++ b/routergroup.go
@@ -201,7 +201,7 @@ func (group *RouterGroup) createStaticHandler(relativePath string, fs http.FileS
 			c.index = -1
 			return
 		}
-		defer f.Close()
+		f.Close()
 
 		fileServer.ServeHTTP(c.Writer, c.Request)
 	}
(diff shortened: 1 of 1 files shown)


routergroup.go
```diff
@@ -193,13 +193,15 @@ func (group *RouterGroup) createStaticHandler(relativePath string, fs http.FileS
...
-		if _, err := fs.Open(file); err != nil {
+		f, err := fs.Open(file)
+		if err != nil {
 			c.Writer.WriteHeader(http.StatusNotFound)
 			c.handlers = group.engine.noRoute
 			// Reset index
 			c.index = -1
 			return
 		}
+		defer f.Close()
```
thinkerou: doesn't use `defer`

Shamus03: Good point, I’ll change that

PR #2120, opened 2019-11-01
Flush any written headers on Hijack

### LINKED ISSUE #2108: What is the purpose of WriteHeaderNow?
Why does gin deviate from net/http in that WriteHeader doesn't actually write the header?

You have to call WriteHeaderNow()

https://github.com/gin-gonic/gin/blob/094f9a9105f8e7b971a01a64677eef5a1f7bcd9b/response_writer.go#L69

This is causing https://github.com/nhooyr/websocket/issues/166

I'll add a fix in my library to call it but I would prefer if gin deprecated this behaviour or added some docs explaining why its necessary.

nhooyr:
Closes #2108

Will make gin compatible with nhooyr.io/websocket.

### DIFF
--- a/response_writer.go
+++ b/response_writer.go
@@ -101,6 +101,13 @@ func (w *responseWriter) Written() bool {
 
 // Hijack implements the http.Hijacker interface.
 func (w *responseWriter) Hijack() (net.Conn, *bufio.ReadWriter, error) {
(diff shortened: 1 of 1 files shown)


nhooyr:
Someone else feel free to pick this up.

PR #2126, opened 2019-11-06
feat: use ShouldBindWith  after ShouldBindBodyWith

mohuishou:
If I use ShouldBindBodyWith in the middleware it will cause me to not be able to use methods like ShouldBindWith that require the Body

- With pull requests:

### DIFF
--- a/context.go
+++ b/context.go
@@ -5,6 +5,7 @@
 package gin
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


context.go
```diff
@@ -666,19 +666,15 @@ func (c *Context) ShouldBindWith(obj interface{}, b binding.Binding) error {
...
-		}
-	}
-	if body == nil {
-		body, err = ioutil.ReadAll(c.Request.Body)
-		if err != nil {
-			return err
-		}
-		c.Set(BodyBytesKey, body)
+	var buf bytes.Buffer
+	tee := io.TeeReader(c.Request.Body, &buf)
```
fifsky: This results in frequent buffer reads, which leads to lower efficiency. If ShouldBindBodyWith is used, subsequent code should only use ShouldBindBodyWith continuously, without sacrificing performance for convenience

mohuishou: You're right if you call ShouldBindBodyWith frequently

PR #2217, opened 2020-01-22
Renew README to fit the modification of SetCookie method

### LINKED ISSUE #2214: SetCookie broken
Hello

It seems you have made a breaking change 4 hours ago (our CI broke)

In source file, SetCookie expects 8 parameters
https://github.com/gin-gonic/gin/blob/master/context.go#L778

In the README, SetCookie expects 7 parameters
https://github.com/gin-gonic/gin/blob/master/README.md
```go
c.SetCookie("gin_cookie", "test", 3600, "/", "localhost", false, true)
```

On the CI
```
... not enough arguments in call to c.SetCookie
	have (string, string, number, string, string, bool, bool)
	want (string, string, int, string, string, http.SameSite, bool, bool)
```

Could you please update the README so that it is consistent with the source code? Otherwise, it is hard to find out where the problem comes from, and which one is the good prototype :-)

From commit f94406a087079bfa5a924cc8095ba753160e885c

Best

panjf2000:
fix #2214

### DIFF
--- a/README.md
+++ b/README.md
@@ -2025,7 +2025,7 @@ func main() {
 
         if err != nil {
             cookie = "NotSet"
-            c.SetCookie("gin_cookie", "test", 3600, "/", "localhost", false, true)
+            c.SetCookie("gin_cookie", "test", 3600, "/", "localhost", http.SameSiteLaxMode, false, true)
         }
 
         fmt.Printf("Cookie value: %s \n", cookie)


panjf2000:
BTY, incompatible changes like that might breaks lots of codebase, maybe `functional options` could be considered to be the solution for those constructions with many optional arguements, just a suggestion.

PR #2316, opened 2020-04-07
Handle invalid Accept header values properly

bonprosoft:
`NegotationFormat` raises out-of-range access to an offer string with some invalid HTTP Accept header values. It happens with the current master: 3315353c20ab224d9308db9df19d83383dba539a .

For example, the following header causes an internal server error.

```
Accept: application/json2
```

I also found `NegotationFormat` returns an inappropriate result with some invalid Accept values. Please refer to the `TestContextNegotiationFormatWithInvalidAccept`.

So I updated the implementation to handle those invalid values properly.

---

I found that the wildcard support for content negotiation was introduced in https://github.com/gin-gonic/gin/pull/1112.
But it seems there is no discussion about the old implementation.
I am sorry if I missed something.

### DIFF
--- a/context.go
+++ b/context.go
@@ -1047,19 +1047,27 @@ func (c *Context) NegotiateFormat(offered ...string) string {
 		return offered[0]
 	}
 	for _, accepted := range c.Accepted {
+		// According to RFC 2616, media-range = ( "*/*" | ( type "/" "*" ) | ( type "/" subtype ) )
+		acceptedMediaRange := strings.Split(accepted, "/")
+		if len(acceptedMediaRange) != 2 || (acceptedMediaRange[0] == "*" && acceptedMediaRange[1] != "*") {
+			continue
+		}
+
 		for _, offer := range offered {
 			// According to RFC 2616 and RFC 2396, non-ASCII characters are not allowed in headers,
-			// therefore we can just iterate over the string without casting it into []rune
-			i := 0
-			for ; i < len(accepted); i++ {
-				if accepted[i] == '*' || offer[i] == '*' {
-					return offer
-				}
-				if accepted[i] != offer[i] {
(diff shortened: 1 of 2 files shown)


rickb777 (changes requested):
This PR does not finish the necessary work. Each media-range can include a quality factor, e.g. `text/xml;q=0.5` and these alter the preference order.

The algorithm should be:
* split the header on `','` into a list of media-ranges.
* foreach media-range, split on `';'` into parts
* split the first part on `'/'` as done in this PR
* find the `q` quality option from the remaining parts; if absent the default is `q=1`
* sort all the parts by quality, without altering the order of those that have the same `q`
* match the sorted list against the offered content

Note that the RFC allows other options as well as `q`.

rickb777:
As it happens, there's a bunch of existing code to parse `Accept` headers such as my own
https://pkg.go.dev/github.com/rickb777/negotiator@v0.5.0#Accept

bonprosoft:
Thank you for your comment.
I'm really sorry for the delay in my response.

My intention of the PR is to prevent gin from having out-of-range access when an invalid HTTP accept header is given.
If someone sends another PR to implement RFC 7231 correctly, then I will close this PR.

rickb777:
I am hoping to have the time to create a patch based on https://pkg.go.dev/github.com/rickb777/negotiator@v0.5.0#Accept, although that code needs to reach v1 first (a bit more testing effort needed there).

rickb777:
I'm still developing https://pkg.go.dev/github.com/rickb777/negotiator, which has changed a lot since I last posted. It's now v0.9.0 and even now not yet stable. However, it _is_ able to work with Gin, which is a good thing.

daheige (approved):

PR #2327, opened 2020-04-18
add `isValidPORTEnvVar` function to judge the  PORT  whether is valid

bestgopher:
If an addr parameter is not passed in `Run`, it will get `PORT` env variable. If there is not a PORT variable，it will use default port - `8080`. But sometimes, there is a PORT env variable, but it is not a valid value(0-65535),the app will panic. So I add a function to judge the PORT whether is valid. For example:

my code:
```
func main() {
    r := gin.Default()
    ...
    if  err := r.Run(); err != nil {
       ...
   }
}
```

run:
```
➜  gin $ echo $PORT
abc
➜  gin $ go run main.go
[GIN-debug] [WARNING] Creating an Engine instance with the Logger and Recovery middleware already attached.

[GIN-debug] [WARNING] Running in "debug" mode. Switch to "release" mode in production.
 - using env:	export GIN_MODE=release
 - using code:	gin.SetMode(gin.ReleaseMode)

[GIN-debug] GET    /get                      --> main.main.func1 (3 handlers)
[GIN-debug] Environment variable PORT="abcd"
[GIN-debug] Listening and serving HTTP on :abcd
[GIN-debug] [ERROR] listen tcp: lookup tcp/abcd: nodename nor servname provided, or not known
2020/04/16 23:33:36 listen tcp: lookup tcp/abcd: nodename nor servname provided, or not known
exit status 1
```

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -11,6 +11,7 @@ import (
 	"path"
 	"reflect"
 	"runtime"
+	"strconv"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


daheige (approved):

utils.go
```diff
@@ -138,14 +139,23 @@ func resolveAddress(addr []string) string {
...
 	case 1:
 		return addr[0]
 	default:
 		panic("too many parameters")
 	}
 }
+
+// Determine the PORT environment variable whether is valid。
+// If the PORT can be parsed to uint(0-65535)，return true。
+func isValidPORTEnvVar(portString string) bool {
```
thinkerou, line 158: use `isValidPortEnvVar`?

daheige: @bestgopher @thinkerou The function is no problem. The function naming can indeed be changed, such as this isValidPortEnvVar Would be better

thinkerou: why use uint or uint16, not uint32?

PR #2339, opened 2020-04-25
modify: nil instead of []string{}

bestgopher:
I think it is better to replace `[] string {}` with `nil` when` key` does not exist, because the zero value of `[] string` type is` nil`.

### DIFF
--- a/context.go
+++ b/context.go
@@ -426,7 +426,7 @@ func (c *Context) GetQueryArray(key string) ([]string, bool) {
 	if values, ok := c.queryCache[key]; ok && len(values) > 0 {
 		return values, true
 	}
-	return []string{}, false
+	return nil, false
 }
 
 // QueryMap returns a map for a given query key.


context.go
```diff
@@ -426,7 +426,7 @@ func (c *Context) GetQueryArray(key string) ([]string, bool) {
 	if values, ok := c.queryCache[key]; ok && len(values) > 0 {
 		return values, true
 	}
-	return []string{}, false
+	return nil, false
```
thinkerou, line 429: please consider:
```
v, ok := GetQueryArray(key)
if len(v) == 0 { // error!
}

**Comment by liuming-dev** [comment]
It seems a broken change!!!

**Comment by bestgopher** [comment]
```
package main

import "fmt"

func test() []string {
	return nil
}

func main() {

	fmt.Println(len(test()))                                  // output：0
	fmt.Println(len(test()) == 0)                         // output：true
	fmt.Println(len(test()) == len([]string{}))    // output：true

	for i, v := range test() {  // no panic
		fmt.Println(i, v)
	}
}
```
@thinkerou If a `[]string` is a `nil`, it's length is `zero` too.

bestgopher:
@liuming-dev Could you tell me the reason?

jokerYellow:
`nil` is `nil`,`[]string{}` is not `nil`.
`[]string{}` is a slice struct with length = 0,capacity = 0,zeroElementAddress = nil.
`nil` is just nil.

daheige (approved):

PR #2341, opened 2020-04-27
Update gin_integration_test.go

ronaldpetty:
TestUnixSocket fails if you run it twice in a row.  This is due to the unix socket file persisting.  Added defer to clean up.

Whats unclear is the following test TestBadUnixSocket I suspect is just looking for cruft maybe from a prior test or defaults not working, I have not enough background to say.

- With pull requests:


ronaldpetty:
Noob question, how do I squash a PR?  I can't seem to find a clear guide for that.

PR #2371, opened 2020-05-12
chore: update tree

thinkerou:
```
if !cond {
}
// else code
```
instead of
```
if cond {
} else {
}
```


thinkerou:
@appleboy I don't know why coverage diff failed.

appleboy (approved):

PR #2379, opened 2020-05-17
Fix the issue of returning validation error instead of json.SyntaxError for a required file upload form field

Alireza-Ta:
If you have a `required` upload file field inside your HTML form, when you send that form with empty file field the binding return an error like so:
```go
&json.SyntaxError{msg:"unexpected end of JSON input", Offset:0}
```
while it must return a validation error!

Simply this PR will fix it!


vsfdqfpdpg (commented):
Fix my problem

PR #2412, opened 2020-06-15
Prevent panic in Context.GetQuery() when there is no Request

pofl:
I have an endpoint that uses an optional query parameter via DefaultQuery(). 
In one unit test case I want to test the route with that parameter.
I write that test as
```go
	w := httptest.NewRecorder()
	c, _ := gin.CreateTestContext(w)
	routeHandler(c)
```
And this panics when routeHandler() calls c.DefaultQuery() because c.Request == nil.

- With pull requests:


pofl:
I have nothing to add to this PR anymore 😊

PR #2455, opened 2020-08-06
delete const MINEXML2

deepoil:
Hi,
This PR,
Deleted the constant "MINEXML2".
Because, according to [RFC7303](
https://tools.ietf.org/html/rfc7303),
text/xml and application/xml are the same, 
and It was written that is recommended to use application/xml to avoid confusion with past usage.
Therefore, I think it is better to remove it from the constants that users can use.

### DIFF
--- a/context.go
+++ b/context.go
@@ -29,7 +29,6 @@ const (
 	MIMEJSON              = binding.MIMEJSON
 	MIMEHTML              = binding.MIMEHTML
 	MIMEXML               = binding.MIMEXML
-	MIMEXML2              = binding.MIMEXML2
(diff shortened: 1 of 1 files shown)


thinkerou:
@DeathKing thanks! maybe we should pilot user not use it, not delete it.

deepoil:
> @DeathKing thanks! maybe we should pilot user not use it, not delete it.

Thank you for confirmation.  
....Why is the reply to another user?  I'll close it for now.

thinkerou:
@deepoil sorry, I use `auto completion` but not check, sorry again.

PR #2484, opened 2020-08-29
Allow bind with a map[string]string

ItalyPaleAle:
It's always been possible to bind using an object of type `map[string]string`. For example:

```go
obj := make(map[string]string)
c.Bind(&obj)
```

If the request body was formatted as JSON (or YAML), this worked fine. You'd get a dictionary with key-values. For example, with input:

```json
{
    "foo": "bar",
    "hello": "world"
}
```

You'd get in the object:

- `obj["foo"] = "bar"`
- `obj["hello"] = "world"`

However, this did **not** work if the request's body was form-encoded (`application/x-www-form-urlencoded` or `multipart/form-data`).

This PR adds support for passing a pointer to `map[string]string` to `Bind` that works with: Forms (POST Forms, Multipart forms), JSON, YAML. It also works with query strings. (However, it does NOT work with XML, because the XML parser doesn't allow that).

**Why is this useful?** Sometimes, there are situations in which I'm expecting a key-value dictionary, but I do not know what keys I'll receive.

This PR includes tests.


ItalyPaleAle:
Hi is there any update on this? @thinkerou or someone else?

ItalyPaleAle:
Ping on this 😄 @thinkerou

binding/form_mapping.go
```diff
@@ -349,3 +364,25 @@ func head(str, sep string) (head string, tail string) {
...
+		}
+	} else {
+		ptrMap, ok := ptr.(map[string]string)
+		if !ok {
+			return errors.New("cannot convert to map of strings")
+		}
+		for k, v := range form {
+			ptrMap[k] = v[len(v)-1] // pick last
+		}
+	}
```
thinkerou, line 386: suggest:
```
if cond {
    // do something
    return
}
// do other something
return
```

thinkerou (approved):
lgtm

appleboy (approved):

PR #2508, opened 2020-09-21
reduce allocs and improve the render `WriteString`

imxyb:
bencmp result

```
benchmark                     old ns/op     new ns/op     delta
BenchmarkReader_Render-12     581           545           -6.20%

benchmark                     old allocs     new allocs     delta
BenchmarkReader_Render-12     9              8              -11.11%

benchmark                     old bytes     new bytes     delta
BenchmarkReader_Render-12     1008          992           -1.59%
```


imxyb:
@appleboy PTAL, thanks

appleboy (approved):

thinkerou (approved):
LGTM

PR #2512, opened 2020-09-24
update content-disposition header to MIME-style

eudore:
Content-Disposition in the `Context.FileAttachment` method is not a MIME style, but a slight performance response that can be avoided.

benchmark test file: [https://github.com/eudore/eudore/blob/master/_example/textprotoHeader_test.go](https://github.com/eudore/eudore/blob/master/_example/textprotoHeader_test.go)


thinkerou (approved):
lgtm

PR #2517, opened 2020-10-01
fix #2505: make Recovery respect gin.DisableConsoleColor

### LINKED ISSUE #2505: Recovery middleware does not obey gin.DisableConsoleColor()
##  Description

Recovery middleware does not obey `gin.DisableConsoleColor()`, but instead has hard coded ansi color sequences always included in its output

## How to reproduce

```
package main

import (
	"fmt"
	"github.com/gin-gonic/gin"
)

func main() {
	gin.DisableConsoleColor()
	g := gin.Default()
	g.GET("/panic", func(c *gin.Context) {
		panic(fmt.Errorf("demo"))
	})
	g.Run(":9000")
}
```

## Expectations

```
$ curl http://localhost:9000/panic
```

Expect logged panic output to not have color codes.

## Actual result

Logged panic has color codes.

## Environment

- go version: 1.14.x
- gin version (or commit ref): 1.6.3
- operating system: Linux

## Additional Information

https://github.com/gin-gonic/gin/blob/master/recovery.go#L54 -- this should be using `gin.IsOutputColor` to determine whether to include color codes in the logger prefix

This is more important when sending the recovery log output through something like `zerolog`, where the color codes just end up being escape codes inside some other string and have no hope of being rendered.

tkgalk:
This addresses #2505. Since Recovery doesn't provide the same config as Layout, this is a naive approach with unexporting ConsoleColorMode variable.

If that's not appropriate I can work over the weekend on `RecoveryWithLogger` which would make Recovery use Logger internally. That would mean however a hard dependency between Recovery and Logger parts.

If there is any other approach you'd like me to pursue I'm open to feedback. :)

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -33,7 +33,7 @@ const (
 	reset   = "\033[0m"
 )
 
-var consoleColorMode = autoColor
+var ConsoleColorMode = autoColor
 
 // LoggerConfig defines the config for Logger middleware.
 type LoggerConfig struct {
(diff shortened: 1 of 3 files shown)


daheige (approved):
LGTM

PR #2552, opened 2020-11-11
Fixes to the graceful shutdown example

averageflow:
…ce before the if statement on graceful shutdown

- With pull requests:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1794,7 +1794,7 @@ func main() {
 	// it won't block the graceful shutdown handling below
 	go func() {
 		if err := srv.ListenAndServe(); err != nil && errors.Is(err, http.ErrServerClosed) {
-			log.Fatalf("listen: %s\n", err)
+			log.Printf("listen: %s\n", err)
 		}
 	}()
 
--- a/README.md
+++ b/README.md
@@ -1793,7 +1793,7 @@ func main() {
 	// Initializing the server in a goroutine so that
 	// it won't block the graceful shutdown handling below
 	go func() {
-		if err := srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+		if err := srv.ListenAndServe(); err != nil && errors.Is(err, http.ErrServerClosed) {
 			log.Fatalf("listen: %s\n", err)
 		}
 	}()
@@ -1812,6 +1812,7 @@ func main() {
 	// the request it is currently handling
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
+	
 	if err := srv.Shutdown(ctx); err != nil {
 		log.Fatal("Server forced to shutdown:", err)
 	}


averageflow:
As stated in the `Shutdown` function of `http.Server`: 
```
// When Shutdown is called, Serve, ListenAndServe, and
// ListenAndServeTLS immediately return ErrServerClosed. Make sure the
// program doesn't exit and waits instead for Shutdown to return.
```
In the graceful shutdown example, the program would exit with `log.Fatalf` thus preventing the graceful shutdown.

averageflow:
A project as Gin that is so loved, should really have some more active maintainers.. Can anyone give a comment on this?

hooman96:
gin needs additional wrapper for graceful shutdown! Currently, we only have `Run()` but there should be an additional function like `RunGracefully()` for such cases.

thinkerou (approved):
lgtm, thanks!

tarupo:
@averageflow Sorry to border you, I just curious about why use ```errors.Is(...)``` but not using ```!errors.Is(...)```. Isn't it will ignore other errors? For example, I started 2 services on the same port(like 8080). And I will not get any errors message printed out, 2 services are still running but no errors are printed. It just looks like nothing going wrong. I think this is kind of weird.
```go
	go func() {
		if err := srv.ListenAndServe(); err != nil && errors.Is(err, http.ErrServerClosed) {
			log.Printf("listen: %s\n", err)
		}
	}()
````

**Comment by averageflow** [comment]
@tarupo to be quite honest, I have altered the example to fit my needs and have a better solution. To be quite honest, you also should not run things on the same port.

here are my utility functions:

```go
const (
	GracefulShutdownRequestGraceSeconds = 10
)

// ListenAndServeGoRoutine will initiate the listen and serve of the wanted *http.Server and
// respond with a log when the server must be closed.
func ListenAndServeGoRoutine(srv *http.Server) {
	err := srv.ListenAndServe()
	if err != nil && errors.Is(err, http.ErrServerClosed) {
		log.Println("Gracefully closing server..")
	}
}

// TerminationSignalWatcher will wait for interrupt signal to gracefully shutdown
// the server with a timeout of x seconds.
func TerminationSignalWatcher(srv *http.Server) {
	// Make a channel to receive operating system signals.
	quit := make(chan os.Signal, 1)

	signal.Notify(quit, syscall.SIGINT, syscall.SIGTERM, syscall.SIGABRT)
	<-quit
	log.Println("Shutting down server, because of received signal..")

	// The context is used to inform the server it has x seconds to finish
	// the request it is currently handling
	ctx, cancel := context.WithTimeout(
		context.Background(),
		GracefulShutdownRequestGraceSeconds*time.Second,
	)

	err := srv.Shutdown(ctx)
	if err != nil {
		log.Fatal("Server forced to shutdown:", err)
	}

	defer cancel()

	log.Println("Server exiting..")
}
```

and how you could call them in your main:

```go
srv := &http.Server{
  ReadTimeout:  10 * time.Second,
  WriteTimeout: 30 * time.Second,
  IdleTimeout:  100 * time.Second,
  Addr:         ":7001",
  Handler:      router,
}

// Initialize the server in a goroutine so that
// it won't block the graceful shutdown handling below
go ListenAndServeGoRoutine(srv)

TerminationSignalWatcher(srv)

```

I feel like this is a better example of how you could get it to work, and can recommend it for everyone. I have this working wonderfully in production for several applications.

averageflow:
@thinkerou maybe you can find some time to add my fonds to the example code and explanation.. i think it is a nicer and cleaner solution. But yes, Go and Gin are already providing you with the puzzle pieces, but i think it is nice for us to showcase it to newcomers

PR #2560, opened 2020-11-16
make it clear that the context will be reset after the request ends, …

jinfeijie:
The `copy()` function cannot forcefully restrict the use of `*gin.context` within the coroutine, so it is easy to cause variable competition. Unconscious things need to be restricted, and the kv carried in a clear context will be recycled after the request ends.


cikay (commented):
The commit message should be "Clarify resetting the context on request end"

Issue #1714, opened 2018-12-25
Reverse Proxy issues


haibeichina:
- With issues:
### router
```go
	apiv1.GET("/videos/:vid-id", ProxyVideoHandler)
	apiv1.POST("/upload/:vid-id", ProxyVideoHandler)
```
### handler
```go
func ProxyVideoHandler(c *gin.Context) {
	u, err := url.Parse(setting.AppSetting.ProxyURL)
	if err != nil {
		logging.Info(err)
		c.String(http.StatusInternalServerError, "internal server error")
		return
	}
	proxy := httputil.NewSingleHostReverseProxy(u)
	c.Request.Host = u.Host
	proxy.ServeHTTP(c.Writer, c.Request)
}
```

- isson is
```shell
[GIN] 2018/12/25 - 12:48:58 | 200 |     15.5698ms |       127.0.0.1 | GET      /static/js/jquery-3.3.1.min.js
[GIN] 2018/12/25 - 12:48:58 | 200 |    20.02673ms |       127.0.0.1 | POST     /api/v1/api
[GIN] 2018/12/25 - 12:48:58 | 200 |   40.379618ms |       127.0.0.1 | POST     /api/v1/api
[GIN] 2018/12/25 - 12:48:58 | 200 |     582.992µs |       127.0.0.1 | GET      /static/img/preloader.jpg


2018/12/25 12:48:58 [Recovery] 2018/12/25 - 12:48:58 panic recovered:
GET /api/v1/videos/973fd869-cba7-4dce-a1c7-8ce2626a1046 HTTP/1.1
Host: 127.0.0.1:8080
Accept: */*
Accept-Encoding: identity;q=1, *;q=0
Accept-Language: zh-CN,zh;q=0.9
Connection: keep-alive
Cookie: session=8a9e75ba-ed79-4c5a-8724-ec3a8e796580; username=hhhhh
Range: bytes=0-
Referer: http://127.0.0.1:8080/api/v1/userhome
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36


net/http: abort Handler
/usr/local/opt/go/libexec/src/runtime/panic.go:513 (0x102bcd8)
	gopanic: reflectcall(nil, unsafe.Pointer(d.fn), deferArgs(d), uint32(d.siz), uint32(d.siz))
[...]

PR #2590, opened 2020-12-23
treat http.ErrAbortHandler as broken pipe too

tmm1:
cc #1714

similar to https://github.com/gorilla/handlers/pull/159/files


m1o1:
Was looking for this as well. Would it make sense to use `errors.Is` instead of using `==` to compare, in case something wraps this error with another error?

PR #2600, opened 2021-01-02
fix compile error from #2572

thinkerou:


thinkerou:
@appleboy hi, we should firstly merge the pull request, because it block other pull request, thanks!

appleboy (approved):

PR #2609, opened 2021-01-12
basic auth: fix timing oracle

### LINKED ISSUE #2226: gin.BasicAuth susceptible to Timing Oracle Attack
## Description
When using gin.BasicAuth to secure a website with only one admin user, the implementation is susceptible to a Timing Oracle Attack.  Such an attack would brute force the password using a dictionary attack, noting that when more characters of the password are properly matched, the call will take longer.  With this scheme each character of the password can be determined given enough time.

Additionally since the application contains the password in plain text, if a intruder is able to make it onto the server, he can use the linux strings command on the binary to yield the stored passwords.  This would mean any intrusion into your server would be a reportable event to regulatory agencies.

## How to reproduce
Please see here for details on the attack: https://en.wikipedia.org/wiki/Timing_attack

## Expectations
1. The code shouldn't require the developer to enter a plain text password to begin with.  Would be better if the the user entered hashed + salted password to begin with.  See linkedin hack for rationale: https://en.wikipedia.org/wiki/2012_LinkedIn_hack
2. For Password comparison, it's best practice to hash or bcrypt the strings prior to comparison to randomize the response time.  Alternatively one could force a rune by rune comparison that always takes constant time (https://golang.org/src/crypto/subtle/constant_time.go).

Snawoot:
Fix #2226 

I've read discussion in #2226, but there are few reasons to consider on behalf of merge of this PR:

* It's pretty easy to fix it
* Even if Basic Auth should not be considered as a secure auth method, it will be often misused in the wild. It'll be hard to make people using basic auth middleware always make informed decision.
* There are cases like server to server integration via HTTPS where basic auth is just good enough.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -5,6 +5,7 @@
 package gin
 
 import (
+	"crypto/subtle"
 	"encoding/base64"
 	"net/http"
 	"strconv"
@@ -30,7 +31,7 @@ func (a authPairs) searchCredential(authValue string) (string, bool) {
 		return "", false
 	}
 	for _, pair := range a {
-		if pair.value == authValue {
+		if subtle.ConstantTimeCompare([]byte(pair.value), []byte(authValue)) == 1 {
 			return pair.user, true
 		}
 	}


thinkerou (approved):
lgtm

PR #2610, opened 2021-01-13
graceful shutdown run function initial setup

hooman96:
Change-Id: I7dca9f7117e54954ad12f5c04f4dd81315008d0e

This PR will add support for graceful shutdown using graceful third party library and expose new function called `RunGracefully()` to the user. 
- tests and complete functionality will be added soon along as either second commit or all squashed commits
- Updated READ.ME to present this new functionality


bestgopher:
How to initialize `Engine .server`  and specify the port？

PR #2621, opened 2021-01-24
engine holds original http server to use all http server functions

ethantsien:
I want to use some original http server methods like func (s *Server) Shutdown(ctx context.Context),so i put the http server in the engine.


ethantsien:
sorry,this PR is not correct.

one engine != one original server

PR #2625, opened 2021-01-29
Mod: move render unittests to render_test package

latavin243:


thinkerou:
@latavin243 thanks, but we should use `render` package, closing.

PR #2642, opened 2021-02-19
adding QueryMapArr

hah:
Adding a new function to context called `QueryMapArray`
at the moment QueryMap only returns the first value from the query params 

example:
let's say this is my query:
`filter[name]=Peter&filter[name]=David`

If do this:
```go
filter := context.QueryMap("filter")
fmt.Println(filter["name"])
```
that would print only the first value (so `Peter`)

QueryMapArray is resolving this, by returning the whole array via `func getarr` instead of just the first value like in `func get`: https://github.com/gin-gonic/gin/blob/1bdf86b722026fd650fddfef7fe9bd8342b51b7a/context.go#L548

Note: maybe there is an existing way to do this, but this one seems a bit more straightforward to me

thanks

### DIFF
--- a/context.go
+++ b/context.go
@@ -463,6 +463,20 @@ func (c *Context) GetQueryMap(key string) (map[string]string, bool) {
 	return c.get(c.queryCache, key)
 }
 
+// QueryMapArray returns a map for a given query key.
+// Same as QueryMap, but the value in the map is an array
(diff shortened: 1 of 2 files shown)


authrequest:
Can we please get this merged?

daheige (approved):

karankohli-cf:
can we get this merged please?

PR #2664, opened 2021-03-27
build: convert to go:build directives

thinkerou:
currently, CI fails on go master branch.

as mentioned in golang/go#41184 and golang/go#44505, Go will probably migrate to a more robust build constraints directive, as explained design doc:

https://golang.org/s/go-build-design

so, we should prepare for that:

for Go-1.17: add all //go:build directives
for Go-1.18: remove all // +build directives

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,34 +18,6 @@ matrix:
     env:
       - TESTTAGS=nomsgpack
   - go: master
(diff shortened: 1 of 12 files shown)


thinkerou:
@appleboy maybe we should review and merge the pr at first, thanks!

appleboy (approved):

PR #2666, opened 2021-03-29
now you can replace binding.Form parser

yanllllk:
- With pull requests:

### DIFF
--- a/binding/binding.go
+++ b/binding/binding.go
@@ -72,17 +72,17 @@ var Validator StructValidator = &defaultValidator{}
 // These implement the Binding interface and can be used to bind the data
 // present in the request to struct instances.
 var (
-	JSON          = jsonBinding{}
-	XML           = xmlBinding{}
-	Form          = formBinding{}
-	Query         = queryBinding{}
-	FormPost      = formPostBinding{}
-	FormMultipart = formMultipartBinding{}
-	ProtoBuf      = protobufBinding{}
-	MsgPack       = msgpackBinding{}
-	YAML          = yamlBinding{}
-	Uri           = uriBinding{}
-	Header        = headerBinding{}
+	JSON          BindingBody = jsonBinding{}
+	XML           BindingBody = xmlBinding{}
+	Form          Binding     = formBinding{}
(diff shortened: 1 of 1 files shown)


binding/binding.go
```diff
@@ -72,17 +72,17 @@ var Validator StructValidator = &defaultValidator{}
...
+	Form          Binding     = formBinding{}
+	Query         Binding     = queryBinding{}
+	FormPost      Binding     = formPostBinding{}
+	FormMultipart Binding     = formMultipartBinding{}
+	ProtoBuf      BindingBody = protobufBinding{}
+	MsgPack       BindingBody = msgpackBinding{}
+	YAML          BindingBody = yamlBinding{}
+	Uri           BindingUri  = uriBinding{}
+	Header        Binding     = headerBinding{}
 )
```
thinkerou, line 87: why add these code?

yanllllk: import (
	"q9admin/framework/gbinding"
	"q9admin/framework/validator"
)

func init() {
	//配置gin的验证器
	binding.Validator = new(validator.DefaultValidator)
      //配置自定义binding解析器,用于自定义类型的binding
     // binding.Form 's type  must be binding.Binding, and you can replace default Form parser 
	binding.Form = gbinding.New()
}




pakage gbinding

type formBinding struct{}

const defaultMemory = 32 << 20

func (formBinding) Name() string {
	return "form"
}

func New() formBinding {
	return formBinding{}
}
.......

func setWithProperType(val string, value reflect.Value, field reflect.StructField) error {
....
	case reflect.Struct:
		switch value.Interface().(type) {
		case time.Time:
			return setTimeField(val, field, value)
		case gtime.Time:  //加入自定义类型的解析,  add new type 
			return setGTimeField(val, field, value)
		}
		return json.Unmarshal([]byte(val), value.Addr().Interface())
...
}


you can replace default Form parser.

yanllllk: or you can add your own type parser..like []int64

TNED:
Is it possible to merge the code here, I want to replace the binding, how to do it

Issue #2669, opened 2021-03-31
Suggestion: Add Method() function to gin.Context to get request method


leungyauming:
## Description

Currently with gin, I need to access `gin.Context.Request.Method` to get the method of the request.

According to the comment of the `Method` field:
`For client requests, an empty string means GET.`
Which means apart from normal method string, there will be empty string too, this will lead to inconsistence of code.

I suggest adding a `Method()` function to `gin.Context` which map empty string to `http.MethodGet`.

PR #2670, opened 2021-03-31
feat: add Method() function to gin.Context

leungyauming:
Relevant Issue: #2669

### DIFF
--- a/context.go
+++ b/context.go
@@ -768,6 +768,15 @@ func (c *Context) IsWebsocket() bool {
 	return false
 }
 
+// Method returns the HTTP method of the requests.
+func (c *Context) Method() string {
+	method := c.Request.Method
+	if method == "" {
+		return http.MethodGet
+	}
+	return method
+}
+
 func (c *Context) requestHeader(key string) string {
 	return c.Request.Header.Get(key)
 }
--- a/README.md
+++ b/README.md
@@ -217,6 +217,29 @@ func main() {
 }
 ```
 
+### Getting Request Method
+
+```go
+package main
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+)
+
+func main() {
+	r := gin.Default()
+
+	r.GET("/", func(c *gin.Context) {
+		c.String(http.StatusOK, fmt.Sprintf("This is a %s request.", c.Method()))
+	})
+
+	r.Run()
+}
+```
+
 ### Parameters in path
 
 ```go
--- a/context_test.go
+++ b/context_test.go
@@ -1764,6 +1764,38 @@ func TestContextShouldBindBodyWith(t *testing.T) {
 	}
 }
 
+func TestContextMethod(t *testing.T) {
+	c, _ := CreateTestContext(httptest.NewRecorder())
+
+	c.Request, _ = http.NewRequest("", "http://example.com", nil)
+	c.Request.Method = ""
+	assert.Equal(t, http.MethodGet, c.Method())
+
+	c.Request, _ = http.NewRequest(http.MethodPost, "http://example.com", nil)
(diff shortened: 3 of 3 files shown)


leungyauming:
```go
func NewRequestWithContext(ctx context.Context, method, url string, body io.Reader) (*Request, error) {
	if method == "" {
		// We document that "" means "GET" for Request.Method, and people have
		// relied on that from NewRequest, so keep that working.
		// We still enforce validMethod for non-empty methods.
		method = "GET"
	}
	if !validMethod(method) {
		return nil, fmt.Errorf("net/http: invalid method %q", method)
	}
	if ctx == nil {
		return nil, errors.New("net/http: nil Context")
	}
	u, err := urlpkg.Parse(url)
	if err != nil {
		return nil, err
	}
	rc, ok := body.(io.ReadCloser)
	if !ok && body != nil {
		rc = io.NopCloser(body)
	}
	// The host's colon:port should be normalized. See Issue 14836.
	u.Host = removeEmptyPort(u.Host)
	req := &Request{
		ctx:        ctx,
		Method:     method,
		URL:        u,
		Proto:      "HTTP/1.1",
		ProtoMajor: 1,
		ProtoMinor: 1,
		Header:     make(Header),
		Body:       rc,
		Host:       u.Host,
	}
	if body != nil {
		switch v := body.(type) {
		case *bytes.Buffer:
			req.ContentLength = int64(v.Len())
			buf := v.Bytes()
			req.GetBody = func() (io.ReadCloser, error) {
				r := bytes.NewReader(buf)
				return io.NopCloser(r), nil
			}
		case *bytes.Reader:
			req.ContentLength = int64(v.Len())
			snapshot := *v
			req.GetBody = func() (io.ReadCloser, error) {
				r := snapshot
				return io.NopCloser(&r), nil
			}
		case *strings.Reader:
			req.ContentLength = int64(v.Len())
			snapshot := *v
			req.GetBody = func() (io.ReadCloser, error) {
				r := snapshot
				return io.NopCloser(&r), nil
			}
		default:
			// This is where we'd set it to -1 (at least
			// if body != NoBody) to mean unknown, but
			// that broke people during the Go 1.8 testing
			// period. People depend on it being 0 I
			// guess. Maybe retry later. See Issue 18117.
		}
		// For client requests, Request.ContentLength of 0
		// means either actually 0, or unknown. The only way
		// to explicitly say that the ContentLength is zero is
		// to set the Body to nil. But turns out too much code
		// depends on NewRequest returning a non-nil Body,
		// so we use a well-known ReadCloser variable instead
		// and have the http package also treat that sentinel
		// variable to mean explicitly zero.
		if req.GetBody != nil && req.ContentLength == 0 {
			req.Body = NoBody
			req.GetBody = func() (io.ReadCloser, error) { return NoBody, nil }
		}
	}

	return req, nil
}
```

As the function called by `http.NewRequest` will auto map empty method string into "GET" which leads a part of the new function not being covered by test, so I turn this pull request into a draft first.

context.go
```diff
@@ -768,6 +768,15 @@ func (c *Context) IsWebsocket() bool {
 	return false
 }
 
+// Method returns the HTTP method of the requests.
+func (c *Context) Method() string {
+	method := c.Request.Method
+	if method == "" {
+		return http.MethodGet
```
BoynChan, line 775: why we should return http.MethodGet when c.Request.Method is a empty string?

leungyauming: image

Because according to the comments of the Method field of the http.Request struct, empty string of Method means GET.

kszafran: Isn't that documentation talking only about "client requests", i.e. requests that you send, not those that you receive?

leungyauming: > Isn't that documentation talking only about "client requests", i.e. requests that you send, not those that you receive?

Seems like it's my understanding problem. Thank you for that.

PR #2675, opened 2021-04-08
fix: data race with trustedCIDRs (#2674)

fifsky:
The RemoteIP function does not have to do the prepareTrustedCIDRs() every time, because engine.Run done it once at startup. This is only for unit testing when loading trusted IP

So you only need to manually execute the PrepareTrustedCIDRS ()  in the unit test


thinkerou (approved):
lgtm, thanks!

appleboy (approved):

yiranzai:
this pr causes the `engine.prepareTrustedCIDRS` to never execute if gin is not started via `gin.RUN()`

PR #2685, opened 2021-04-11
RouterGroup.Handle regular match optimization of http method

daheige:
RouterGroup.Handle matches the optimization of http method. There is no need to do regular matching of reg every time. If there are multiple grouping routes, it will consume more performance.


chaoyueziji (approved):

thinkerou (approved):
lgtm

appleboy (approved):

PR #2692, opened 2021-04-15
set engine.TrustedProxies For items that don't use gin.RUN

yiranzai:
#2632 

https://github.com/gin-gonic/gin/pull/2675

#2675 causes the `engine.prepareTrustedCIDRS` to never execute if `gin` is not started via `gin.RUN()`.

this pr set `engine.TrustedProxies` and execute `engine.prepareTrustedCIDRS` For items that don't use `gin.RUN`

Our project uses `http.Handler` to start the service instead of `gin.Run`

### DIFF
--- a/context_test.go
+++ b/context_test.go
@@ -1392,14 +1392,10 @@ func TestContextAbortWithError(t *testing.T) {
 	assert.True(t, c.IsAborted())
 }
 
-func resetTrustedCIDRs(c *Context) {
-	c.engine.trustedCIDRs, _ = c.engine.prepareTrustedCIDRs()
-}
-
 func TestContextClientIP(t *testing.T) {
 	c, _ := CreateTestContext(httptest.NewRecorder())
 	c.Request, _ = http.NewRequest("POST", "/", nil)
-	resetTrustedCIDRs(c)
+	c.engine.trustedCIDRs, _ = c.engine.prepareTrustedCIDRs()
 	resetContextForClientIPTests(c)
 
 	// Legacy tests (validating that the defaults don't break the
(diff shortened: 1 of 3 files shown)


yiranzai:
The failed is not my reason https://travis-ci.org/github/gin-gonic/gin/jobs/767120791

gin.go
```diff
@@ -366,6 +366,23 @@ func (engine *Engine) prepareTrustedCIDRs() ([]*net.IPNet, error) {
 	return cidr, nil
 }
 
+// SetTrustedProxies  set Engine.TrustedProxies
+func (engine *Engine) SetTrustedProxies(trustedProxies []string) error {
```
thinkerou, line 370: please add some unit test, thanks!

yiranzai: done

context_test.go
```diff
@@ -1428,57 +1424,47 @@ func TestContextClientIP(t *testing.T) {
...
-	c.engine.TrustedProxies = []string{"40.40.25.25/16", "30.30.30.30"}
-	resetTrustedCIDRs(c)
+	_ = c.engine.SetTrustedProxies([]string{"40.40.25.25/16", "30.30.30.30"})
 	assert.Equal(t, "20.20.20.20", c.ClientIP())
 
 	// Use hostname that resolves to all the proxies
-	c.engine.TrustedProxies = []string{"foo"}
-	resetTrustedCIDRs(c)
+	_ = c.engine.SetTrustedProxies([]string{"foo"})
 	assert.Equal(t, "40.40.40.40", c.ClientIP())
```
yiranzai, line 1449: I passed the previous unit tests, but I don't understand why the `foo` `bar` `bar` would expect such a result

gin.go
```diff
@@ -326,11 +326,11 @@ func iterate(path, method string, routes RoutesInfo, root *node) RoutesInfo {
 func (engine *Engine) Run(addr ...string) (err error) {
 	defer func() { debugPrintError(err) }()
 
-	trustedCIDRs, err := engine.prepareTrustedCIDRs()
+	err = engine.parseTrustedProxies()
```
estroz, line 329: Why not add this to all `Run*` methods too?

yiranzai: Because the other `RUN*` didn't exist before.

My core idea was to be able to dynamically adjust the TurstedProxies.

yiranzai: Fix bugs caused by #2675 

The bugs:

https://github.com/gin-gonic/gin/issues/2693

https://github.com/gin-gonic/gin/issues/2697

xpunch:
TrustedProxies is set with []string{"0.0.0.0/0"} by default, you'd better give trustedCIDRs a default value. So we don't need to call SetTrustedProxies when we don't need to change it.

yiranzai:
> TrustedProxies is set with []string{"0.0.0.0/0"} by default, you'd better give trustedCIDRs a default value. So we don't need to call SetTrustedProxies when we don't need to change it.

I haven't changed any of the previous logic, this is a PR that avoids https://github.com/gin-gonic/gin/pull/2675 bug in a new way. As for what you said, `gin.Run` did that

xpunch:
Your code is used to set custom TrustedProxies when don't use gin.Run.
I think the simplest way is set default value for trustedCIDRs. 
```go
func New() *Engine {
	...
        engine.trustedCIDRs = []*net.IPNet{{IP: net.IP{0x0, 0x0, 0x0, 0x0}, Mask: net.IPMask{0x0, 0x0, 0x0, 0x0}}}
        ...
}
```
Default value of trustedCIDRs is missing when gin.New() called, TrustedProxies is set to []string{"0.0.0.0/0"}, which means trustedCIDRs should also set a default value matchs TrustedProxies.

evanfuller:
@yiranzai what is the status on this change? I would love to use it.

atcharles:
The easiest way:
Copy the following to add to the reverse proxy:
proxy_set_header X-Appengine-Remote-Addr $remote_addr;

FeodorFitsner:
When this fix is planned for release?

thinkerou (approved):
lgtm, thanks!

appleboy (approved):

leafduo:
This fix has 2 problems:
1. it's not backward-compatible 
2. is doesn't update related documents

almas1992:
Is there any plan to release this change?

yveshield:
> The easiest way:
> Copy the following to add to the reverse proxy:
> proxy_set_header X-Appengine-Remote-Addr $remote_addr;
thanks.

rbeuque74:
Hello,

> Is there any plan to release this change?

I couldn't agree more. Our software have been bumped to v1.7 because we follow security advisories, and you published https://github.com/advisories/GHSA-h395-qcrw-5vmq. But at this time, without the release of `SetTrustedProxies`, we can't use the new version in production.
Would you mind to release `v1.7.5` to fix that ?

Thanks,
Romain

PR #2708, opened 2021-04-28
Optimization reflect

daheige:
reflect code maping optimize


daheige:
Hope to optimize the reflection here, thank you

daheige:
ok

PR #2716, opened 2021-05-09
Check multipart file header size on test

y-yagi:
`FileHeader.Size` was added in Go 1.9. So we can use it now.
https://golang.org/pkg/mime/multipart/#FileHeader


appleboy (approved):

thinkerou (approved):
lgtm

PR #2723, opened 2021-05-18
Get client IP when using Cloudflare

ItalyPaleAle:
When proxying an app through Cloudflare, the user's real IP is set in the CF-Connecting-IP header. This is more convenient than having to manually maintain an allow-list of trusted proxies.

Reference: https://support.cloudflare.com/hc/en-us/articles/200170986-How-does-Cloudflare-handle-HTTP-Request-headers-

### DIFF
--- a/context.go
+++ b/context.go
@@ -731,8 +731,13 @@ func (c *Context) ShouldBindBodyWith(obj interface{}, bb binding.BindingBody) (e
 // If the headers are nots syntactically valid OR the remote IP does not correspong to a trusted proxy,
 // the remote IP (coming form Request.RemoteAddr) is returned.
 func (c *Context) ClientIP() string {
-	if c.engine.AppEngine {
+	switch {
+	case c.engine.AppEngine:
 		if addr := c.requestHeader("X-Appengine-Remote-Addr"); addr != "" {
+			return addr
+		}
+	case c.engine.CloudflareProxy:
+		if addr := c.requestHeader("CF-Connecting-IP"); addr != "" {
 			return addr
 		}
 	}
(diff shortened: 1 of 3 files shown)


appleboy (approved):

ItalyPaleAle:
Thinking about it, I wonder if rather than using booleans, it might make more sense to create an enum `platform` (or something like that)?

This way, once the next platform is inevitably added, we don't add yet another boolean.

The downside is around what to do with the existing Google App Engine flag.

thinkerou:
> Thinking about it, I wonder if rather than using booleans, it might make more sense to create an enum `platform` (or something like that)?
> 
> This way, once the next platform is inevitably added, we don't add yet another boolean.
> 
> The downside is around what to do with the existing Google App Engine flag.

Yes, I also like `enum`

ItalyPaleAle:
I can update to use an enum-like.

What should I do for the existing Google App Engine flag? Keep that for legacy and "translate" that into the enum?

thinkerou (approved):
lgtm

ItalyPaleAle:
Wait before you merge.... I want to update it to use an enum. Before we introduce even more legacy. :) just let me know what I should do with the existing Boolean flag for Google App Engine

thinkerou:
> I can update to use an enum-like.
> 
> What should I do for the existing Google App Engine flag? Keep that for legacy and "translate" that into the enum?

@ItalyPaleAle I merged the pull request at first, please commit other pull request using(refactoring) `enum`, thanks!

ItalyPaleAle:
Ok. Will do it over the weekend or sooner.

What do you want me to do with the existing Boolean for GAE? I can't remove it because it would be backwards-incompatible. I can keep it as a legacy alias that then is translated to the enum?

thinkerou:
@ItalyPaleAle please read https://github.com/gin-gonic/gin/blob/master/deprecated.go for reference, we should deprecated warning at first and remove.

ItalyPaleAle:
@thinkerou I made the change in #2739 hope this works!

PR #2739, opened 2021-05-28
Setting trusted platform using an enum-like

ItalyPaleAle:
Updated #2723 using an enum-like.

This also marks `AppEngine` as deprecated (note that we can't detect people using `AppEngine: false` because that's also the zero value, so we can only show an error message to people who set `AppEngine: true`).

This removes the `Cloudflare` property set in #2723, which was never released anyways.

### DIFF
--- a/context.go
+++ b/context.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
 	"math"
 	"mime/multipart"
 	"net"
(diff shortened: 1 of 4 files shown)


ItalyPaleAle:
@thinkerou this was the modified PR you requested. Let me know if it's good or you want me to make further changes

thinkerou (approved):
lgtm

thinkerou:
waiting @appleboy review, thanks!

appleboy (approved):

PR #2776, opened 2021-07-05
Use bit shift operation instead of division

zihengCat:
- Use bit shift operation instead of division in `context.go`.
- Add constant comments.


appleboy (approved):

thinkerou (approved):
lgtm

PR #2782, opened 2021-07-08
Use std http method constant instead of raw string

zihengCat:
- Use std `net/http` http method constant instead of raw string in `gin.go`.


thinkerou (approved):
lgtm

PR #2784, opened 2021-07-10
update go.mod

daheige:
update go.mod package


thinkerou:
@daheige thanks a lot! #2827 have updated, please see it.

PR #2785, opened 2021-07-11
export ErrUnknownType and code adjust for broken pipe and json invalid and use http.MethodGet instead of raw GET string and IsWebsocket check

daheige:
1.export ErrUnknownType unknown request type
2.code adjust for broken pipe and json invalid and use http.MethodGet instead of raw GET string
3.IsWebsocket check

### DIFF
--- a/binding/json.go
+++ b/binding/json.go
@@ -13,16 +13,21 @@ import (
 	"github.com/gin-gonic/gin/internal/json"
 )
 
-// EnableDecoderUseNumber is used to call the UseNumber method on the JSON
-// Decoder instance. UseNumber causes the Decoder to unmarshal a number into an
-// interface{} as a Number instead of as a float64.
-var EnableDecoderUseNumber = false
+var (
+	// EnableDecoderUseNumber is used to call the UseNumber method on the JSON
+	// Decoder instance. UseNumber causes the Decoder to unmarshal a number into an
+	// interface{} as a Number instead of as a float64.
+	EnableDecoderUseNumber = false
 
-// EnableDecoderDisallowUnknownFields is used to call the DisallowUnknownFields method
-// on the JSON Decoder instance. DisallowUnknownFields causes the Decoder to
-// return an error when the destination is a struct and the input contains object
-// keys which do not match any non-ignored, exported fields in the destination.
-var EnableDecoderDisallowUnknownFields = false
+	// EnableDecoderDisallowUnknownFields is used to call the DisallowUnknownFields method
+	// on the JSON Decoder instance. DisallowUnknownFields causes the Decoder to
+	// return an error when the destination is a struct and the input contains object
+	// keys which do not match any non-ignored, exported fields in the destination.
+	EnableDecoderDisallowUnknownFields = false
+
+	// JsonReqInvalid json invalid when request is nil or request body is nil.
+	JsonReqInvalid = errors.New("invalid request")
+)
 
 type jsonBinding struct{}
 
@@ -32,7 +37,7 @@ func (jsonBinding) Name() string {
 
 func (jsonBinding) Bind(req *http.Request, obj interface{}) error {
 	if req == nil || req.Body == nil {
-		return errors.New("invalid request")
+		return JsonReqInvalid
 	}
 	return decodeJSON(req.Body, obj)
 }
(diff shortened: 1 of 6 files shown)


daheige:
@appleboy @thinkerou

binding/json.go
```diff
@@ -13,16 +13,21 @@ import (
...
+	EnableDecoderUseNumber = false
+
+	// EnableDecoderDisallowUnknownFields is used to call the DisallowUnknownFields method
+	// on the JSON Decoder instance. DisallowUnknownFields causes the Decoder to
+	// return an error when the destination is a struct and the input contains object
+	// keys which do not match any non-ignored, exported fields in the destination.
+	EnableDecoderDisallowUnknownFields = false
+
+	// JsonReqInvalid json invalid when request is nil or request body is nil.
+	JsonReqInvalid = errors.New("invalid request")
```
appleboy, line 29: Maybe `InvaildRequest`?

daheige:
The name you said is OK, but the JSON request request is illegal. I'll adjust it appropriately.


------------------ Original ------------------
From: Bo-Yi Wu ***@***.***&gt;
Date: Tue,Jul 13,2021 9:46 AM
To: gin-gonic/gin ***@***.***&gt;
Cc: heige ***@***.***&gt;, Author ***@***.***&gt;
Subject: Re: [gin-gonic/gin] export ErrUnknownType and code adjust for broken pipe and json invalid and use http.MethodGet instead of raw GET string and IsWebsocket check (#2785)





 
@appleboy commented on this pull request.
 
 
In binding/json.go:
 &gt; -// keys which do not match any non-ignored, exported fields in the destination. -var EnableDecoderDisallowUnknownFields = false +var ( +// EnableDecoderUseNumber is used to call the UseNumber method on the JSON +// Decoder instance. UseNumber causes the Decoder to unmarshal a number into an +// interface{} as a Number instead of as a float64. +EnableDecoderUseNumber = false + +// EnableDecoderDisallowUnknownFields is used to call the DisallowUnknownFields method +// on the JSON Decoder instance. DisallowUnknownFields causes the Decoder to +// return an error when the destination is a struct and the input contains object +// keys which do not match any non-ignored, exported fields in the destination. +EnableDecoderDisallowUnknownFields = false + +// JsonReqInvalid json invalid when request is nil or request body is nil. +JsonReqInvalid = errors.New("invalid request")  
Maybe InvaildRequest?

binding/json.go
daheige: OK, I've adjusted it.

recovery.go
```diff
@@ -47,6 +47,20 @@ func RecoveryWithWriter(out io.Writer, recovery ...RecoveryFunc) HandlerFunc {
 	return CustomRecoveryWithWriter(out, defaultHandleRecovery)
 }
 
+// IsBroken Check error for a broken connection
+func IsBroken(err interface{}) bool {
```
thinkerou, line 51: public method?

daheige:
yes，I see that other components of gin are in use. Here, it is declared as an open method to facilitate reuse.


------------------ Original ------------------
From: thinkerou ***@***.***&gt;
Date: Thu,Jul 22,2021 8:00 AM
To: gin-gonic/gin ***@***.***&gt;
Cc: heige ***@***.***&gt;, Author ***@***.***&gt;
Subject: Re: [gin-gonic/gin] export ErrUnknownType and code adjust for broken pipe and json invalid and use http.MethodGet instead of raw GET string and IsWebsocket check (#2785)





 
@thinkerou commented on this pull request.
 
 
In recovery.go:
 &gt; @@ -47,6 +47,20 @@ func RecoveryWithWriter(out io.Writer, recovery ...RecoveryFunc) HandlerFunc {  return CustomRecoveryWithWriter(out, defaultHandleRecovery)  }   +// IsBroken Check error for a broken connection +func IsBroken(err interface{}) bool {  
public method?

binding/json.go
```diff
@@ -32,7 +37,7 @@ func (jsonBinding) Name() string {
 
 func (jsonBinding) Bind(req *http.Request, obj interface{}) error {
 	if req == nil || req.Body == nil {
-		return errors.New("invalid request")
+		return InvalidRequest
```
kszafran, line 40: How about:
```suggestion
		return ErrInvalidRequest
```
I think it's a common way to name errors.

daheige: ok,thinks @kszafran

appleboy:
@daheige Please help to resolve conflicts.

PR #2812, opened 2021-08-04
feat: add methods `c.BindCookie(obj)` from cookie and `c.BindRequest(obj)` from http request

tangx:
- With pull requests:

---

1. split `Bind` method into 2 parts
  + 1. extract  binding code as `BindOnly()` / `BindBodyOnly()` method, in this way, users can combine their own binding logic.
  + 2. `Bind()` / `BindBody()` method handle error from `BindOnly` / `BindBodyOnly`, and do validation stuff.

```go
type Binding interface {
	Name() string
	Bind(*http.Request, interface{}) error
	BindOnly(*http.Request, interface{}) error
}

type BindingBody interface {
	Binding
	BindBody([]byte, interface{}) error
	BindBodyOnly([]byte, interface{}) error
}


```

2. add `c.BindCookie(obj) method, which  can bind values from request cookie into object. 

3. add `c.BindRequest(obj)` method, which combines `uri, form , cookie, header, body` handlers. With this method, now can bind all wanted values at once.

```go
// ...
type Params struct {
	Name          string `uri:"name"`
	Age           int    `form:"age,default=18"`
	Money         int32  `form:"money" binding:"required"`
	Authorization string `cookie:"Authorization"`
	UserAgent     string `header:"User-Agent"`
	Data          struct {
		Replicas *int32 `json:"replicas" yaml:"replicas" xml:"replicas" form:"replicas"`
	} `body:"body"`
}


// ### POST test by RestClient of vscode extenstion
// POST http://127.0.0.1:9881/hello/zhangsan?money=1000
// Content-Type: application/json
// Accept-Language: en-GB,en-US;q=0.8,en;q=0.6,zh-CN;q=0.4
// Cookie: Authorization=auth123123;
//
// {
//     "replicas":5
// }




```

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Gin is a web framework written in Go (Golang). It features a martini-like API wi
     - [Bind Uri](#bind-uri)
     - [Bind Header](#bind-header)
     - [Bind Cookie](#bind-cookie)
-	- [Bind Request](#bind-request)
+    - [Bind Request](#bind-request)
     - [Bind HTML checkboxes](#bind-html-checkboxes)
     - [Multipart/Urlencoded binding](#multiparturlencoded-binding)
     - [XML, JSON, YAML and ProtoBuf rendering](#xml-json-yaml-and-protobuf-rendering)
@@ -1003,9 +1003,13 @@ see `Authorization` in `Params` of [Bind Request](#bind-request)
 
 `c.BindRequest` and `c.ShouldBindRquest` can bind all wanted values into struct instance at once.
 Now, 
-+ supports values from `form`, `uri`, `header`, `cookie` 
-and request `req.Body`.
++ supports values from `query`, `uri`, `header`, `cookie`, `form` 
+and `body` for request `req.Body`.
 + `req.Body` decocder is decided by header `Content-Type` value, see more [request_test.go](./binding/binding_request_test.go#L40)
+
+> ❗️ Attention: 
+>> 1. tag `query`, `uri`, `header`, `cookie` must never be used in body struct, or it will be panic.
+>> 2. tag `query` only work for `c.BindRequest` and `c.ShouldBindRequest` api.
 
 **example**
 
@@ -1019,8 +1023,8 @@ import (
 
 type Params struct {
 	Name          string `uri:"name"`
-	Age           int    `form:"age,default=18"`
-	Money         int32  `form:"money" binding:"required"`
+	Age           int    `query:"age,default=18"`
+	Money         int32  `query:"money" binding:"required"`
 	Authorization string `cookie:"Authorization"`
 	UserAgent     string `header:"User-Agent"`
 	Data          struct {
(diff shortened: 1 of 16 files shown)


quentinvernot:
Hello,

Not a gin maintainer, just a random passer-by here.

First off, I love this API, I implemented a similar thing as a wrapper but it makes sense for gin to expose this directly. However, there is a small but fairly dangerous issue here: gin (and encoding/json) will try to guess the name of query fields on untagged struct fields, meaning if you have a struct like this:
```go
type MyRequest struct {
  MyField string `query:"my_field"`
}
```

... all the binders will try to fill this field in the order they are called. Not just `BindQuery`, `BindHeaders` and `BindJSON` will try it too.

If you send this request:
```
curl -X POST https://myserver/my-handler?my_field=query_value -d '{"my_field": "json_value"}'
```
... and bind with `BindRequest`, the `MyField` field of the `MyRequest` object will contain `"json_value"`.

This also applies to fields that are not tagged at all, which can be dangerous security-wise. If you have, say, a `UserID` field that you would expect a middleware to fill, it could be filled by `BindRequest` and possible overwrite a value, easy way in for an attacker.

There are ways around this:
- A lot of `reflect` acrobatics to create temporary sub structs for each tag, you can then pass the sub-struct to the relevant `BindX` function and then merge everything back into the expected struct to aggregate the results.
- Add an option in gin to stop it from guessing field names, but beware that this wouldn't apply to `json` or `xml` because that's how the go standard library works.
- Document the issue

[Echo](https://echo.labstack.com/guide/binding/) has a similar binding function and does 2 and 3.

Just my 2 cents.

Cheers!

tangx:
> gin (and encoding/json) will try to guess the name of query fields on untagged struct fields, meaning if you have a struct like this

this is cause by using a same function `mapping` to bind values, in `mapping` has a recursive mapping behavior in struct field.

---

another problem, gin has no `query` tag. `c.BindQuery` api is using tag `form`, so it's hard to divided them in body struct.
so that , I add my own binding query method in `requestBinding` object to support `query` tag , and it can only be used in `c.ShouldBindRequest` api.

by the way,  it will be panic if body struct contains any of `query, uri, cookie, header`

try follow branch

```bash
go mod edit -replace=github.com/gin-gonic/gin=github.com/tangx/gin@develop
```

long2ice:
That's what I want

long2ice:
Why there is `Data` struct?

long2ice:
I update your branch like this and work fine, any other points?
image

tangx:
> Why there is `Data` struct?

more info in https://github.com/tangx/gin#bind-request

1. `Data` sturct is  used for binding  datas from `resp.Body` only.
2. `Data` struct  can be any name , but must be with the `body:""` tag. 
3.  there is only one `resp.Body` in request, so it can have only one `body` tag in your `Params' instance`


```
// ### POST test by RestClient of vscode extenstion
// POST http://127.0.0.1:9881/hello/zhangsan?money=1000
// Content-Type: application/json
// Accept-Language: en-GB,en-US;q=0.8,en;q=0.6,zh-CN;q=0.4
// Cookie: Authorization=auth123123;
//
// {
//     "replicas":5
// }
```

tangx:
try to use this to bind request

https://github.com/tangx/ginbinder

PR #2841, opened 2021-08-22
Group types and vars

alexander-melentyev:


alexander-melentyev:
@appleboy @thinkerou

thinkerou:
@alexander-melentyev thanks a lot, in order to reduce code indent that easier to read.

lancoLiu:
I think the original way is also readable

PR #2852, opened 2021-09-02
Add go-clean-template

alexander-melentyev:


alexander-melentyev:
@appleboy @thinkerou

daheige (changes requested):
disagree it

PR #2893, opened 2021-10-06
fix bug : gin config `RemoveExtraSlash` do not work when  use static …

goodaye:
when i  config static path in my project 

code: 
```
	app = gin.Default()
	app.RemoveExtraSlash = true
	app.Static("/static", "my-static-path")
```
request
```
# curl 127.0.0.1/static/test.json
# test.json  

# curl 127.0.0.1///static/test.json
# 404 page not found
```

because  handler `createStaticHandler.func1` ignore `rPath` ,  so  change   request data  then handler will process request in right way . 

```
c.Request.URL.Path = rPath
```

- With pull requests:

### DIFF
--- a/gin.go
+++ b/gin.go
@@ -519,6 +519,7 @@ func (engine *Engine) handleHTTPRequest(c *Context) {
 
 	if engine.RemoveExtraSlash {
 		rPath = cleanPath(rPath)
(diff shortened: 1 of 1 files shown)


thinkerou:
please add some unit test, thanks!

PR #2978, opened 2021-12-04
fix:param is empty string not response 404.

codeHauler-1:
fix:param is empty string not response 404.


Fov6363 (changes requested):
need add test

PR #2989, opened 2021-12-17
Add GetPostFormData to return c.formCache

Awezome:
as titled


jincheng9:
What's the purpose?

Awezome:
> What's the purpose?

some time we need to print all post form data in log file .
I use the code below to get post form data now , but I think it is very ugly .

`
c.GetPostFormMap("")
fmt.Println(c.Request.PostForm)
`

PR #3015, opened 2022-01-05
Removing $ as when you copy the command $ causes an issue in terminal.

chowmean:
- Readme Change: Removed the $ sign as when you copy the command it doesn't work in terminal.


jincheng9:
need to be discussed. `$` is commonly used in the docs.

chowmean:
I think $ was used to indicate shell. Next commands doesn't have this.

Bisstocuz:
$ was used to indicate shell, and you can also remove it when you type in command.

chowmean:
Makes sense for the shell. But when I start using it I just copied to the whole command and it didn't work the first time as I have to remove the $. Closing this as this doesn't seem relevant.

PR #3030, opened 2022-01-20
Fix logging of IPv6 addresses with scope

eric:
The `net.ParseIP()` function does not handle IPv6 addresses with a scope (with `%en0` at the end, for instance).

It may make sense to move to something with better IPv6 support like https://pkg.go.dev/inet.af/netaddr in the future.

See https://tailscale.com/blog/netaddr-new-ip-type-for-go/ for more details.


Bisstocuz:
Add some test cases plz

PR #3051, opened 2022-02-06
test: add TestBindingDefault test case

MichaelDeSteven:
* default case should include MIMEHTML case
* TestBindingDefault function needs consider contentType is MsgPack or MIMEHTML case


MichaelDeSteven:
the error fix link https://github.com/gin-gonic/gin/pull/3050
plz fix ASAP

Issue #2919, opened 2021-10-27
Can i bind uri and query to a struct in one method?


zhaojinxin409:
## Description

Can i bind uri and query in one method? Currently when i bind uri , it will return an error because of binding of query is empty.

Maybe we need a `BindUrl` function ?

## How to reproduce

```
package main

import (
	"github.com/gin-gonic/gin"
	"log"
)

type DatasourceListRequest struct {
	Name  string `uri:"name" binding:"required"`
	Query string `form:"query" binding:"required""`
}

func main() {
	g := gin.Default()
	g.GET("/hello/:name", Handler)
	g.Run(":9000")
}

func Handler(ctx *gin.Context) {
	var req = new(DatasourceListRequest)
	if err := ctx.ShouldBindUri(req); err != nil {
		log.Println("bind uri error")
	}
	if err := ctx.ShouldBindQuery(req); err != nil {
		log.Println("bind query error")
	}
}
```

zhaojinxin409:
> > Maybe you are right,
> > > Maybe we need a `BindUrl` function ?
> > 
> > 
> > Or we can simply let `BindUri` support query binding. Query field is empty when checking the struct after binding URI.
> 
> `BindUrl` may be a better solution because of backwards compatibility :)

Any idea about it? i can try to contribute if the idea is accepted

ggicci:
What about creating a middleware to achieve this? and considering a third-party package like [ggicci/httpin](https://ggicci.github.io/httpin/integrations/gin)

hongnguyenhuu96:
I found that there is a method `MapFormWithTag` in the binding package
for example
if you want to skip binding query and only bind uri and validate after bind uri

```go
        var req serializers.InternalGetMaxExpressCODAmountReq
	if err := binding.MapFormWithTag(&req, c.Request.URL.Query(), "form"); err != nil {
		log.Err(err).Ctx(c).Msg("Failed to parse request form")
		response.Error(c, http.StatusBadRequest, err, translate.ErrorBadRequest)
		return
	}

	if err := c.ShouldBindUri(&req); err != nil { // call bind only one time
		log.Err(err).Ctx(c).Msg("Failed to parse request uri")
		response.Error(c, http.StatusBadRequest, err, translate.ErrorBadRequest)
		return
	}```

weiyinfu:
Now that gin doesn't support BindUrl()，we should think about other solutions.   
Maybe we can define the struct better.
[...]

PR #3095, opened 2022-03-22
Add a shortcut to bind uri and query at the same time

pnck:
- Imagine URL like `.../person/:name/resources?order=1`
- And we have structs like
```golang
type Person struct {
  Name string `uri:"name"`
}
type LookupControl struct {
 Order SomeEnum `form:"order"`
}
```
-  I would like to bind both of them without checking error twice:
```golang
var req struct { 
 Person
 LookupControl 
}
// prefer
if err := c.ShouldBindUrl(&req); err != nil { // ...

// rather than
if err := c.ShouldBindUri(&req.Person); err != nil {
 logError(err)
}
if err := c.ShouldBindQuery(&req.LookupControl); err != nil {
 logError(err)
}
```

It's discussed in #2919 early before but seems no PR presented.

### DIFF
--- a/context.go
+++ b/context.go
@@ -647,6 +647,14 @@ func (c *Context) BindUri(obj any) error {
 	return nil
 }
 
+// BindUrl combines c.BindUri() and c.BindQuery(), abort with HTTP 400 likewise
+func (c *Context) BindUrl(obj any) error {
+	if err := c.BindUri(obj); err != nil {
+		return err
+	}
+	return c.BindQuery(obj)
+}
+
 // MustBindWith binds the passed struct pointer using the specified binding engine.
 // It will abort the request with HTTP 400 if any error occurs.
 // See the binding package.
@@ -702,6 +710,16 @@ func (c *Context) ShouldBindUri(obj any) error {
 		m[v.Key] = []string{v.Value}
 	}
 	return binding.Uri.BindUri(m, obj)
+}
+
+// ShouldBindUrl combines c.ShouldBindQuery() and c.ShouldBindUri().
+// It will bind uri parameters and query part at the same time.
+func (c *Context) ShouldBindUrl(obj any) error {
+	err := c.ShouldBindUri(obj)
+	if err != nil {
+		return err
+	}
+	return c.ShouldBindQuery(obj)
 }
 
 // ShouldBindWith binds the passed struct pointer using the specified binding engine.
--- a/githubapi_test.go
+++ b/githubapi_test.go
@@ -285,7 +285,7 @@ var githubAPI = []route{
 	{http.MethodDelete, "/user/keys/:id"},
 }
 
-func TestShouldBindUri(t *testing.T) {
+func TestShouldBindUrl(t *testing.T) {
 	DefaultWriter = os.Stdout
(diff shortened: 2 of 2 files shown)


pticptenec:
Hello, @pnck!
I also came up with need of this feature. Great that you made the PR.

I think there's need in this, because:
1)validator package(gin dependency) supports cross-binding out of the box
2)query parameters are part of the URIs

so it's odd that BindUri method doesn't bind Query parameters.

But I think, there should be some improvements on this PR
(although I'm not a maintainer).
1)URL is a subset of URI's, (here's [link](https://stackoverflow.com/questions/176264/what-is-the-difference-between-a-uri-a-url-and-a-urn) question with refers to wiki and RFC) so I think it would be convenient if the original method
BindUri will do the same as yours BindUrl
2)...But I think in that case it will be a breaking change with MustBind
group of methods (if they start emit errors in client's code with custom structures...)
3)I looked up in the repository and I think appropriate place
for your tests is not a githubapi_test.go which contains whole application
but context_test.go.

One of the possible solutions it's use
combination with method ShouldBindWith and provide appropriate binding
...but in that case URI is a special binding which needs a context.Params
data, but others don't.

So, honestly, I don't see a way except a breaking change with Must/ShouldBindUri methods.
Or just leave ShouldBind without Must companion method.

pnck:
> Hello, @pnck! I also came up with need of this feature. Great that you made the PR.
> 
> I think there's need in this, because: 1)validator package(gin dependency) supports cross-binding out of the box 2)query parameters are part of the URIs
> 
> so it's odd that BindUri method doesn't bind Query parameters.
> 
> But I think, there should be some improvements on this PR (although I'm not a maintainer). 1)URL is a subset of URI's, (here's [link](https://stackoverflow.com/questions/176264/what-is-the-difference-between-a-uri-a-url-and-a-urn) question with refers to wiki and RFC) so I think it would be convenient if the original method BindUri will do the same as yours BindUrl 2)...But I think in that case it will be a breaking change with MustBind group of methods (if they start emit errors in client's code with custom structures...) 3)I looked up in the repository and I think appropriate place for your tests is not a githubapi_test.go which contains whole application but context_test.go.
> 
> One of the possible solutions it's use combination with method ShouldBindWith and provide appropriate binding ...but in that case URI is a special binding which needs a context.Params data, but others don't.
> 
> So, honestly, I don't see a way except a breaking change with Must/ShouldBindUri methods. Or just leave ShouldBind without Must companion method.


Oh I've even forgotten this PR.  Much appreciated for your reply.

Honestly neither do I see that much necessity of adding new APIs. Feature like so can be easily achieved by composing a trivial function with existing bindings. And I guess that's the reason why #2919 had laid there for a while but maintainers didn't pay their attentions.

I came up with this PR originally intended to be a proposal or supplement of #2919 and hoped to see if maintainers have had any ideas. So I may spend not much effort until there are new changes or "official statements" comming out. But your advices are very valuable though, pretty sure that should be considered in further proposals (mine or others', if this feature actually adopted).

jurevito:
Hello!
How come this is not a thing already? Seems very counter intuitive to have two structs for each GET request.

Issue #2334, opened 2020-04-21
Validation issue


dimuska139:
## Description

There is no way to handle validation error If struct field has type "int" and user sends string value.

## How to reproduce

```go
package main

import (
	"fmt"
	"github.com/gin-gonic/gin"
	"github.com/go-playground/validator/v10"
	"net/http"
        "reflect"
)

func main() {
	g := gin.Default()
	g.GET("/articles", func(c *gin.Context) {
		type Paginator struct {
			Page int `form:"page" binding:"required,min=1"`
		}
		var pag Paginator
		if err := c.ShouldBindQuery(&pag); err != nil {
			errs := err.(validator.ValidationErrors)
			respErrors := make(map[string]interface{})
			for _, v := range errs {
				field, _ := reflect.TypeOf(&pag).Elem().FieldByName(v.Field())
				fieldName, _ := field.Tag.Lookup("form")
				respErrors[fieldName] = v.Tag()
			}
			c.JSON(http.StatusBadRequest, gin.H{"errors": respErrors})
			return
		}
		c.String(200, "Articles list is here")
	})
	g.Run(":9000")
}
```

## Expectations

For example: ```/articles?page=1``` is correct but ```/articles?page=stringishere``` throws panic. How to resolve this situation and return correct error message ```{"errors":"page":"Number expected"}```? 

## Environment

- go version: go1.12.7 linux/amd64
- gin version (or commit ref): 1.6.2
- operating system: Linux Mint

canercidam:
@dimuska139 The error comes from type assertion line:

```go
errs := err.(validator.ValidationErrors)
```

```
interface conversion: error is *strconv.NumError, not validator.ValidationErrors
```

You probably need to do something like:

```go
errs, ok := err.(validator.ValidationErrors)
if !ok {
	c.String(400, "Handle differently")
	return
}
```

or this, which is even better handling IMO:

```go
switch err.(type) {
case validator.ValidationErrors:
	respErrors := make(map[string]interface{})
	for _, v := range err.(validator.ValidationErrors) {
		field, _ := reflect.TypeOf(&pag).Elem().FieldByName(v.Field())
		fieldName, _ := field.Tag.Lookup("form")
		respErrors[fieldName] = v.Tag()
	}
	c.JSON(http.StatusBadRequest, gin.H{"errors": respErrors})

case *strconv.NumError:
	c.String(400, "Handle differently")
}
```

dimuska139:
[...]

PR #3097, opened 2022-03-22
feat: introduce binding.FieldError type to tell the erroneous field

maxatome:


maxatome:
Seems to close #2334.

PR #3100, opened 2022-03-24
Fix some tests

mstmdev:
1. Sleep for one millisecond in the handler because the `Latency` will return `0s` sometimes and the test will fail.
2. The `TCPListener.File` is not supported by windows, it is unimplemented now.
3. Remove the `LF` in the `testdata/template/raw.tmpl`, because if set the git config `core.autocrlf=true`, will append `CR` to the `raw.tmpl` automatically, then test is failed on Windows.


daheige (approved):

thinkerou (approved):
lgtm

PR #3156, opened 2022-05-27
Multiple `Accept` header fix

tsmethurst:
This PR updates Accept content negotiation to iterate through multiple Accept headers to find possible negotiable formats, rather than just taking the first Accept header and ignoring the others.

### DIFF
--- a/context.go
+++ b/context.go
@@ -1100,7 +1100,9 @@ func (c *Context) NegotiateFormat(offered ...string) string {
 	assert1(len(offered) > 0, "you must provide at least one offer")
 
 	if c.Accepted == nil {
-		c.Accepted = parseAccept(c.requestHeader("Accept"))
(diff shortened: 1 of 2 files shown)


tsmethurst:
Hello! Can this be merged? It's still pretty useful imo :)

daenney:
@thinkerou @appleboy Would it be possible to get some eyes on this one?

It's very easy to run into this situation. When using the Go and preparing a request people might easily call `req.Header.Add("Accept", ...)` multiple times instead of concatenating the values with a `,`, and on the command line people are very likely to use `-H "Accept: ..."` with cURL instead of building the string by hand.

Based on my reading of the RFC, any field that accepts a comma separated list can also be represented as the same field multiple times. Given Accept is a list, Gin should understand both.

PR #3183, opened 2022-06-09
fix: Nil pointer dereference when context has no engine defined

ismdeep:
Issue: #3178


jbleduigou:
Thanks for the PR 🙏 
It seems similar to https://github.com/gin-gonic/gin/pull/3179

PR #3186, opened 2022-06-09
Cookies added to return all cookies at once.

arshamalh:
http.Request has two methods called Cookie and Cookies, I think it's a good option for Gin to have both of them, not limiting them to only one method.
without this commit, user should call cookies one by one, handle errors (cookie doesn't exist) and user can't find out what other cookies are there.
Also Express that is gin inspiration can handle multiple cookies at once.

### DIFF
--- a/context.go
+++ b/context.go
@@ -897,6 +897,16 @@ func (c *Context) Cookie(name string) (string, error) {
 	return val, nil
 }
 
+// Cookies parses and returns the HTTP cookies sent with the request.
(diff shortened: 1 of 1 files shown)


arshamalh:
http.Request has two methods called Cookie and Cookies, I think it's a good option for Gin to have both of them, not limiting them to only one method.
without this commit, user should call cookies one by one, handle errors (cookie doesn't exist) and user can't find out what other cookies are there.

PR #3188, opened 2022-06-10
Prevent slice out of bounds access before panic

### LINKED ISSUE #3126: Wildcard conflict encounter unexpected runtime.boundsError
## Description

**master branch**

At tree.go:355, there is `pathSeg := strings.SplitN(n.children[0].path, "/", 2)[0]`, and `n.children[0]` will cause runtime.boundsError.

## How to reproduce

```
package main

import (
	"github.com/gin-gonic/gin"
)

func main() {
	g := gin.Default()
	g.GET("/src1/", func(c *gin.Context) {
		c.String(200, "Hello World!")
	})
        g.GET("/src1/*filepath", func(c *gin.Context) {
		c.String(200, "Hello World!")
	})
	g.Run(":9000")
}
```

## Expectations

run code `panic("catch-all wildcard '" + path +
				"' in new path '" + fullPath +
				"' conflicts with existing path segment '" + pathSeg +
				"' in existing prefix '" + n.path + pathSeg +
				"'")`

## Actual result

encounter runtime.boundsError at `pathSeg := strings.SplitN(n.children[0].path, "/", 2)[0]`

## Environment

- go version: 1.18
- commit ref: a06d546f5c2e853b07fde5a41b1fff102829b116
- operating system: Linux

fivemoreminix:
Previously, when the error would be triggered, it was possible that a slice out of bounds panic would occur before the intended panic message. I've added a nil/slice length check to make sure that does not happen.

With that said, I am not very familiar with the library so I cannot determine what this catch was supposed to mean, or convey, to users of the library. It would be helpful to know what I can do to make the error message more useful.

Thanks.

Fixes #3126

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -349,7 +349,11 @@ func (n *node) insertChild(path string, fullPath string, handlers HandlersChain)
 		}
 
 		if len(n.path) > 0 && n.path[len(n.path)-1] == '/' {
-			pathSeg := strings.SplitN(n.children[0].path, "/", 2)[0]
+			// Wildcard conflict
+			pathSeg := path
+			if len(n.children) != 0 {
+				pathSeg = strings.SplitN(n.children[0].path, "/", 2)[0]
+			}
 			panic("catch-all wildcard '" + path +
 				"' in new path '" + fullPath +
 				"' conflicts with existing path segment '" + pathSeg +


reza-ebrahimi:
Any update on this?

Root route is not working with wildcard routes:

```
GET /api/auth => not works
GET /api/auth/*session
```

PR #3232, opened 2022-07-08
fix(readme): graceful shutdown should use buffered signal

rickwang7712:
According to [Notify document](https://pkg.go.dev/os/signal#Notify), the channel used with signal.Notify should be buffered. Otherwise, there will be a risk that the signal might be missed.
Linters will also report this problem:
sigchanyzer: misuse of unbuffered os.Signal channel as argument to signal.Notify (govet)go-golangci-lint
SA1017: the channel used with signal.Notify should be buffered (staticcheck)go-golangci-lint

- With pull requests:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1841,7 +1841,7 @@ func main() {
 
 	// Wait for interrupt signal to gracefully shutdown the server with
 	// a timeout of 5 seconds.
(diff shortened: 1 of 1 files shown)


rickwang7712:
duplicated with #3220

Issue #3241, opened 2022-07-12
[BUG]:  index out of range [8] with length 8


secsys-go:
- With issues:
  - Use the search tool before opening a new issue.
  - Please provide source code and commit sha if you found a bug.
  - Review existing issues and provide feedback or react to them.

## How to reproduce
Poc is here:

```
package main

import (
	"fmt"
	"net/http"
	"net/http/httptest"

	"github.com/gin-gonic/gin"
)

func main() {
	c, _ := gin.CreateTestContext(httptest.NewRecorder())
	c.Request, _ = http.NewRequest("POST", "/", nil)
	c.Request.Header.Add("Accept", "text/html")
	fmt.Println(c.NegotiateFormat("text/htm"))
}

```

## Expectations

It prints "" and exit without error

## Actual result
It  crashed with:

```
panic: runtime error: index out of range [8] with length 8

goroutine 1 [running]:
github.com/gin-gonic/gin.(*Context).NegotiateFormat(0xc000261000, {0xc00032ff60, 0x1, 0x6})
        /home/zjx/workspace/gowork/src/go-fdg-exmaples/gin/context.go:1110 +0x186
main.main()
        /home/zjx/workspace/gowork/src/go-fdg-exmaples/gin/pocTestContextNegotiation/main.go:14 +0x1db
exit status 2
```

## Environment

- go version:
go version go1.17.8 linux/amd64
- gin version (or commit ref):
gin with commit c4580944ae8edba96809660ffbfdadc6caaefc2d
- operating system:
ubuntu 18.04 64bit

stareezy-1:
Root cause: `Context.NegotiateFormat` looped `for i < len(accepted)` and indexed `offer[i]` without checking `i < len(offer)`. When an accepted format is longer than the offered one and shares a full prefix (e.g. `Accept: text/html` vs `text/htm`, `image/tiff-fx` vs `image/tiff`), the loop reads past the end of the offer slice and panics with `index out of range [8] with length 8`.

This is already fixed on master by #3397 (commit 7cb151b, Jan 2023), which bounds the loop with `i < len(accepted) && i < len(offer)` and adds regression coverage via `TestContextNegotiationFormat2`. I verified the exact reproducer from this issue: it panics on the pre-fix code and returns `""` (the expected result) on current master.

No code changes needed — closing as fixed.

PR #3243, opened 2022-07-12
Fix NegotiateFormat

eleven26:
Fix issue [#3241](https://github.com/gin-gonic/gin/issues/3241)

When the length of the `offered` mime type string is inconsistent with the request's `accepted` mime length, it may cause a panic, but it should actually return an empty string.


thinkerou (approved):

thinkerou:
please also review the pull request @dawkaka thanks!

dawkaka (changes requested):
This works fine.
But we could skip looping through if offer is shorter than accepted and doesn't also end with `*`

Issue #3239, opened 2022-07-11
Not support customize http method！such as lowercase “get”


SaltySooda:
- With issues:
  - Use the search tool before opening a new issue.
  - Please provide source code and commit sha if you found a bug.
  - Review existing issues and provide feedback or react to them.

## Description
Not support customize http method！such as lowercase “get”. 

## How to reproduce

```
package main

import (
	"github.com/gin-gonic/gin"
)

func main() {
	g := gin.Default()
	g.get("/hello/:name", func(c *gin.Context) {
		c.String(200, "Hello %s", c.Param("name"))
	})
	g.Post("/hello/:name", func(c *gin.Context) {
		c.String(200, "Hello %s", c.Param("name"))
	})
	g.UpDaTe("/hello/:name", func(c *gin.Context) {
		c.String(200, "Hello %s", c.Param("name"))
	})
	g.Run(":9000")
}

**Comment by 359859461** [comment]
@thinkerou @eleven26 
？
gin only support some uppercase http method：
`const (
	MethodGet     = "GET"
	MethodHead    = "HEAD"
	MethodPost    = "POST"
	MethodPut     = "PUT"
	MethodPatch   = "PATCH" // RFC 5789
	MethodDelete  = "DELETE"
	MethodConnect = "CONNECT"
	MethodOptions = "OPTIONS"
	MethodTrace   = "TRACE"
)`
So，what if I want to receive“get” request from outside？

**Comment by eleven26** [comment]
> @thinkerou @eleven26 ？ gin only support some uppercase http method： `const ( MethodGet = "GET" MethodHead = "HEAD" MethodPost = "POST" MethodPut = "PUT" MethodPatch = "PATCH" // RFC 5789 MethodDelete = "DELETE" MethodConnect = "CONNECT" MethodOptions = "OPTIONS" MethodTrace = "TRACE" )` So，what if I want to receive“get” request from outside？

Are you saying that gin can't handle `get` requests? If yes, please post your code or http message or other useful information.

**Comment by 359859461** [comment]
```
package main

import (
	"github.com/gin-gonic/gin"
	"net/http"
	"strings"
	"time"
)

func main() {
	//server
	g :=gin.Default()
	g.GET("/hello/:name", func(c *gin.Context) {
		c.String(200, "Hello %s", c.Param("name"))
	})
	go g.Run(":9000")
	time.Sleep(time.Second)
	//client
	client1()
	client2()
	for true {
		time.Sleep(time.Second)
	}
}

[...]

PR #3244, opened 2022-07-13
add new function for creating "customize http method" handler

SaltySooda:
#3239


359859461:
@thinkerou please review,thanks

bestgopher:
I think these is a similar method `RouterGroup.Handle`.

PR #3249, opened 2022-07-18
remove not in use files that have vulnerability

AbhilashVijayakumar:
- we dont use yaml request payload


AbhilashVijayakumar:
sorry should be on fork

Issue #2634, opened 2021-02-08
Variation of the Unix timestamp issue


zettelmj:
Hi,

I am running into a variation of https://github.com/gin-gonic/gin/issues/1979 difference in my case is that the timestamp is in milliseconds. My approach would be to follow https://github.com/ThomasObenaus/gin/commit/ecddd0378397087c6524e6924e4a8995ee215d2a example and extend the supported format to not only support unix and unixnano but also unixmilli.

Would such a PR be welcome?

Thank you
Jens

takanuva15:
I was just looking whether gin can do a binding directly from milliseconds and I'm confused - was this ever implemented? 
Or do we need to keep dividing the timestamp by 1000 to fit the "unix" time_format when submitting requests through gin?

Issue #1979, opened 2019-07-06
binding:extended time.Time type, support Unix time


guonaihong:
I sometimes use createTime (unix time) as the key for data segmentation. A common usage is that Service A is responsible for generating creatTime. Service B routes to the corresponding database table based on createTime in the data.
It is found that gin does not currently support binding of unix time to time.Time type.

Pseudo code is as follows
```golang
package main

import (
        "fmt"
        "github.com/gin-gonic/gin"
        "time"
)

type shareTime struct {
        CreateTime time.Time `form:"createTime" time_format:"unixNano"`
        UnixTime   time.Time `form:"unixTime" time_format:"unix"`
}

func main() {
        r := gin.Default()
        unix := r.Group("/unix")

        testCT := time.Date(2019, 7, 6, 16, 0, 33, 123, time.Local)
        fmt.Printf("%d\n", testCT.UnixNano())

        testUT := time.Date(2019, 7, 6, 16, 0, 33, 0, time.Local)
        fmt.Printf("%d\n", testUT.Unix())

        unix.GET("/nano", func(c *gin.Context) {
                s := shareTime{}

                c.ShouldBindQuery(&s)

                if !testCT.Equal(s.CreateTime) {
                        c.String(500, "want %d got %d", testCT.UnixNano(), s.CreateTime)
                        return
                }

                c.JSON(200, s)
        })

        unix.GET("/sec", func(c *gin.Context) {
                s := shareTime{}

                c.ShouldBindQuery(&s)

                if !testUT.Equal(s.UnixTime) {
                        c.String(500, "want %d got %d", testCT.Unix(), s.UnixTime)
                        return
                }

                c.JSON(200, s)

        })

        r.Run()
}

```

IsQiao:
the feature can't bind requrest json body property. any solution about it?

PR #3257, opened 2022-07-28
feat(binding): add support for unixMilli and unixMicro

peakji:
Related issues: https://github.com/gin-gonic/gin/issues/2634, https://github.com/gin-gonic/gin/issues/1979

This PR adds support for two new `time_format` variants: `unixMilli` and `unixMicro`, the former of which is added to make front-end JavaScript code more convenient to work with.

The implementation is taken from [time.UnixMilli](https://cs.opensource.google/go/go/+/go1.18.4:src/time/time.go;l=1349) and [time.UnixMicro](https://cs.opensource.google/go/go/+/go1.18.4:src/time/time.go;l=1355), which were introduced to the standard library in go 1.17. For compatibility with go 1.15, these two functions are not used directly.

### DIFF
--- a/binding/binding_test.go
+++ b/binding/binding_test.go
@@ -69,15 +69,19 @@ type FooStructDisallowUnknownFields struct {
 }
 
 type FooBarStructForTimeType struct {
-	TimeFoo    time.Time `form:"time_foo" time_format:"2006-01-02" time_utc:"1" time_location:"Asia/Chongqing"`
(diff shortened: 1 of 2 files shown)


takanuva15:
Hey sorry I know this was some ago, but why was the PR closed instead of merged? Is there a new way to directly bind a millisecond timestamp to a form struct with gin?

helenxu1221:
Could this fix be merged please?

PR #3275, opened 2022-08-14
chore: run tests on go 1.19

Aoang:


thinkerou:
@Aoang thanks! but duplicate #3272

PR #3286, opened 2022-08-17
Fix: 404 bug with param

StephanoGeorge:
Code to reproduce the bug:

```go
package main

import "github.com/gin-gonic/gin"

func main() {
	router := gin.Default()
	router.GET("/user/:id/posts", func(c *gin.Context) { c.Status(200) })
	router.GET("/:general/root", func(c *gin.Context) { c.Status(200) })
	router.Run()
}
```

```sh
curl http://localhost:8080/user/root
404 page not found
```


spirit1431007:
same issue

PR #3287, opened 2022-08-17
feat: skip url with prefixes in LoggerConfig

ddzhj:
As in a RESTful url design, resource identities often appear as part of url, such as `/resources/show/:id`. With `SkipPaths` alone in `LoggerConfig`, we can not handle `skipping-logging-by-path` with urls one by one. So here comes the `prefix` way of `skipping-logging-by-prefix-match`. 

Tests delievered.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/mattn/go-isatty"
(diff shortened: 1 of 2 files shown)


raulci:
Hi, I was looking a way of skipping logs with prefixes and arrived here.
I see this PR was not finally merged, @aichaoxy any reason why? Is there another way of doing it without this PR?

I want to skip some URLs like `/assets/css/common.cs` , `/assets/img/banner.jpg`, etc. by just setting `/assets`, but it didn't work.
I can always set all the files, but wondering to make it work with just the prefix.

Thanks.

PR #3331, opened 2022-09-20
Update README.md

Badr008:
this is the first time trying a framework in Go , so when I followed the documentation I found this outdated CLI command block me from installing the library , so here it is the latest command available .
Thank You


oussamabouchareb:
as far as I'm aware "install" command is meant for installing binaries and "get" is meant for installing packages, why would want to install gin as binary?

Jarch09:
Agree with @oussamabouchareb

PR #2769, opened 2021-06-28
fallback Context.Deadline() Context.Done() Context.Err() to Context.Request.Context()

wei840222:
base on comment https://github.com/gin-gonic/gin/pull/2751#issuecomment-867461945 in https://github.com/gin-gonic/gin/pull/2751
also fallback Context.Deadline() Context.Done() Context.Err() to Context.Request.Context()

PR #3367, opened 2022-10-18
Modify c.Request.Context() when using ContextWithFallback

bvisness:
Gin 1.8 added the `ContextWithFallback` feature flag, which causes the `context.Context` methods to fall back to `c.Request.Context()`. Presumably, the intent of this feature was to cause the Gin context to be a part of the overall context chain of the request. However, this intent breaks down when using `gin.WrapH` with a third-party `http.Handler`.

Because the signature of an `http.Handler` does not include a `*gin.Context`, an `http.Handler` can only access `req.Context()`. However, this always refers to the request's original context instead of the Gin context. This means that any values set with Gin's `c.Set` are not accessible in those handlers. This means that Gin middleware that sets context values actually can't pass values to third-party handlers!

This PR therefore sets `c.Request.Context()` to be the Gin context itself. That way, `req.Context()` in a third-party handler returns the Gin context, and can get the values it expects.

It only does this when using `ContextWithFallback`, to avoid breakage. I figure this is probably enough breakage-avoidance, because this flag is already opt-in and most people won't opt in. However, if you needed to avoid further breakage, I could perhaps update this to _only_ do this work when using `gin.WrapH`.

Refs #2751, #2769, #3166, #3172

### DIFF
--- a/context.go
+++ b/context.go
@@ -5,6 +5,7 @@
 package gin
 
 import (
+	"context"
 	"errors"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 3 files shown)


bvisness:
Naturally, minutes after finishing this, I thought of a much less invasive fix for our problem. I have opened #3368 as a result. Closing this.

Issue #3351, opened 2022-10-04
Connection reset by peer


jamesli2021:
## Description
Connection reset by peer always occurred after benchmark.

## How to reproduce

Refer to the source for wrk benchmark
```
https://github.com/PacktPublishing/Building-Distributed-Applications-in-Gin/tree/main/chapter07/api-without-db
```
Change this line `gin.Default()` to `gin.New()`

```
go mod init gintest
go mod tidy
GIN_MODE=release ./e
```

## Expectations

Benchmark should return no stack error
```
wrk -c 100 http://localhost:8080/recipes
```
Large JSON response will retrieve from `/recipes` endpoint.

## Actual result

Error stack will throw right after the benchmark finished.
```
2022/10/04 22:55:10 http: panic serving [::1]:57818: write tcp [::1]:8080->[::1]:57818: write: connection reset by peer
goroutine 30 [running]:
net/http.(*conn).serve.func1()
	/usr/local/go/src/net/http/server.go:1850 +0xb0
panic({0x104c6a5c0, 0x14000530050})
	/usr/local/go/src/runtime/panic.go:890 +0x258
github.com/gin-gonic/gin/render.JSON.Render(...)
	/Users/s/go/pkg/mod/github.com/gin-gonic/gin@v1.8.1/render/json.go:58
github.com/gin-gonic/gin.(*Context).Render(0x14000513d00, 0xc8, {0x104cbe180, 0x140005060b0})
	/Users/s/go/pkg/mod/github.com/gin-gonic/gin@v1.8.1/context.go:910 +0xf4
github.com/gin-gonic/gin.(*Context).JSON(...)
...
```

## Environment

- go version: 1.19.1
- gin version (or commit ref): 1.8.1
- operating system: macOS Ventura
- hardware: Macbook Pro (M1)

jamesli2021:
So this caused panic:
https://github.com/gin-gonic/gin/blob/master/render/json.go#L58

ElegantSoft:
Any updates

xuxin1:
**so how to solve this problem**······🤔

ElegantSoft:
The problem was in nginx

jamesli2021:
It's the problem with the panic, happen without nginx when you change to `router.Run(":3000")` or random ports.

jamesli2021:
> The problem was in nginx

Are you referring to your setup and same codebase?

jamesli2021:
[...]

PR #3390, opened 2022-11-07
Don't panic when failed to render or write JSON

git-hulk:
Currently, gin will panic if failed to write JSON to client and it's unnecessary for most scenarios. Because most of them are caused by network error like connection reset or timeout, it'd be better to retrieve an error instead of panic.

This PR can also fix this issue: https://github.com/gin-gonic/gin/issues/3351

### DIFF
--- a/render/json.go
+++ b/render/json.go
@@ -54,10 +54,7 @@ var (
 
 // Render (JSON) writes data with custom ContentType.
 func (r JSON) Render(w http.ResponseWriter) (err error) {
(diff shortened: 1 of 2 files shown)


danmux (approved):
Whilst this does not change the API signature it does change the behaviour, which may be considered a breaking change, but these are rare and clients of this call should be handling errors.

This also makes the API more consistent with other renderers

thinkerou:
duplicate #2150

PR #3425, opened 2022-12-06
Wrap the context render panic.

danmux:
The utility of the `c.JSON` helper or other `context.Render` methods not returning an error simplifies client code.

There are rare errors during rendering that result in a panic, which most production systems will handle in middleware.

This introduces the ability to detect that it is a gin rendering error in the panic handler as opposed to other handler logic panics.

To be able to trap c.JSON render panics this PR will also need https://github.com/gin-gonic/gin/pull/3390 - which changes the JSON renderer to be more consistent, until that is also merged this context Render panic will not fire.

- With pull requests:

### DIFF
--- a/context.go
+++ b/context.go
@@ -922,7 +922,10 @@ func (c *Context) Render(code int, r render.Render) {
 	}
 
 	if err := r.Render(c.Writer); err != nil {
-		panic(err)
+		panic(Error{
+			Err:  err,
+			Type: ErrorTypeRender,
+		})
 	}
 }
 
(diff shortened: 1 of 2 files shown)


danmux:
I wonder if since this is only used in the json render panic the error should be more specific.
I was thinking that other helpers in future may panic - and it might be annoying to have to introduce error types per renderer

danmux:
better handled by. https://github.com/gin-gonic/gin/pull/2150/files

Issue #3459, opened 2023-01-03
panic with HandleMethodNotAllowed and route corner case


kristiansvalland:
## Description

Setting `HandleMethodNotAllowed` on the engine and configuring the routes in a intricate way, gin panics when a certain request is sent. See details below on how to reproduce.

## How to reproduce

```
package main

import (
	"github.com/gin-gonic/gin"
)

func main() {
	r := gin.Default()
	r.HandleMethodNotAllowed = true

	r.GET("/base/metrics", handler)
	r.GET("/base/v1/:id/devices", handler)
	r.GET("/base/v1/user/:id/groups", handler)
	r.GET("/base/v1/organizations/:id", handler)
	r.DELETE("/base/v1/organizations/:id", handler)
	r.Run() 
}
```

Now, make a `GET` on "localhost:8080/base/v1/user/groups", in which case gin panics.

I must admit that the above setup is a bit off, with how ".../:id/devices" and ".../user/:id/groups" is overlapping somewhat, but I still do not expect a panic. If the setup of the routes truly is wrong, then I would expect a panic to happen earlier, either at `r.GET()` or `r.Run()`.

I've also tried to delete some of the routes to minimize the example, but it seems that all lines are needed for this to fail, which further emphasize that this is a corner case. 

## Expectations

```
$ curl http://localhost:8080/base/v1/user/groups
404 page not found
```

## Actual result

```
$ curl -i http://localhost:8080/base/v1/user/groups
curl: (52) Empty reply from server
```

With the following log message from the gin app:

```
2023/01/03 10:16:50 http: panic serving 127.0.0.1:56181: runtime error: index out of range [0] with length 0
goroutine 34 [running]:
net/http.(*conn).serve.func1()
        /usr/local/go/src/net/http/server.go:1850 +0xb0
panic({0x1053aefe0, 0x140000260f0})
        /usr/local/go/src/runtime/panic.go:890 +0x258
github.com/gin-gonic/gin.(*node).getValue(0x1400012d938?, {0x14000520244?, 0x1400012d9d8?}, 0x0, 0x1400000e030, 0x0)
        /Users/kristiansvalland/go/pkg/mod/github.com/gin-gonic/gin@v1.8.2/tree.go:637 +0x378
github.com/gin-gonic/gin.(*Engine).handleHTTPRequest(0x1400054c4e0, 0x140000ae000)
[...]

PR #3460, opened 2023-01-03
fix(router): tree bug where loop index is not decremented.

kristiansvalland:
Also add test to validate fix.

fixes https://github.com/gin-gonic/gin/issues/3459

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -459,9 +459,9 @@ walk: // Outer loop for walking the tree
 					// If the path at the end of the loop is not equal to '/' and the current node has no child nodes
 					// the current node needs to roll back to last valid skippedNode
 					if path != "/" {
-						for l := len(*skippedNodes); l > 0; {
-							skippedNode := (*skippedNodes)[l-1]
-							*skippedNodes = (*skippedNodes)[:l-1]
+						for length := len(*skippedNodes); length > 0; length-- {
+							skippedNode := (*skippedNodes)[length-1]
+							*skippedNodes = (*skippedNodes)[:length-1]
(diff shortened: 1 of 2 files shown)


tree.go
```diff
@@ -462,6 +462,7 @@ walk: // Outer loop for walking the tree
 						for l := len(*skippedNodes); l > 0; {
```
thinkerou, line 462: use `for l := len(*skippedNodes); l >0; l--)`?

maybe rename `l` to `length`, because `l` like `1`

tree.go
```diff
@@ -579,6 +580,7 @@ walk: // Outer loop for walking the tree
 				for l := len(*skippedNodes); l > 0; {
 					skippedNode := (*skippedNodes)[l-1]
 					*skippedNodes = (*skippedNodes)[:l-1]
+					l--
```
thinkerou, line 583: as above

tree.go
```diff
@@ -636,6 +638,7 @@ walk: // Outer loop for walking the tree
 			for l := len(*skippedNodes); l > 0; {
 				skippedNode := (*skippedNodes)[l-1]
 				*skippedNodes = (*skippedNodes)[:l-1]
+				l--
```
thinkerou, line 641: as above

tree.go
kristiansvalland: Much cleaner. Fixed :)

thinkerou (approved):
lgtm, thanks!

appleboy (approved):

tree.go
beage666: Up date

PR #3471, opened 2023-01-15
feature: write query bool parser for query parameters.

dozheiny:
I added a query bool parser for URL query parameters. This parser will convert query parameters with a boolean value to a Golang bool type, making it easier to work with in the code. Test cases have been added to ensure proper functionality.
Thanks.

### DIFF
--- a/context.go
+++ b/context.go
@@ -15,6 +15,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


alirezaeftekhari (changes requested):
Dear @dozheiny ,
I think 2 tips would make your feature better,

1) Give the client code more flexibility if there is a boolean parse error to handle.
For example, your `QueryBool` method implementation can be like this:
```go
func (c *Context) QueryBool(key string) (bool, error) {
	value, _ := c.GetQuery(key)
	boolValue, err := strconv.ParseBool(value)
	return boolValue, err
}
```
In this way, the client has better flexibility to handle the GET boolean input if there is a parse error. In your implementation, the return value from your `QueryBool` method will be false if a parse error exists and the client does not know this value is an explicit false value or if it is because of a parser error.

2) Most frameworks will calculate the boolean value true if it is a string, you are calculating it false and it may cause misunderstanding for clients that are new in Gin.

dozheiny:
Hi Alireza, your right.
I committed to the changes you asked for.

PR #3479, opened 2023-01-27
feature: add BindForm function

dozheiny:
Adding the BindForm function to the context. This feature will make it much easier for developers to handle form data in their applications. The added flexibility will improve the user experience and make the development process more efficient.

### DIFF
--- a/context.go
+++ b/context.go
@@ -661,6 +661,11 @@ func (c *Context) BindHeader(obj any) error {
 	return c.MustBindWith(obj, binding.Header)
 }
 
+// BindForm is a shortcut for c.MustBindWith(obj, binding.Form)
+func (c *Context) BindForm(obj any) error {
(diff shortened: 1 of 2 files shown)


alirezaeftekhari (dismissed):
Yes, I agree with you,
We Usually use the `Bind` method that dedicates the `content-type` and handles the functionality with that.
But it will be better if we have another method that handles forms explicitly.
It would be better if you also implement the `ShouldBindForm` method and its corresponding test.

dozheiny:
Hi;
Sure, I will commit to the changes you asked for.

dozheiny:
I made the changes you asked for.

alirezaeftekhari (approved):

dozheiny:
Would one of the maintainers be able to review this PR?

Issue #3555, opened 2023-04-01
FileAttachment() is vulnerable to Reflect File Download


motoyasu-saburi:
I have contacted the maintainer 4-5 times via email, Snyk, etc., 
but have not received a reply, so I am raising an Issue.

## Description

`gin` > `context.go` > `func FileAttachment` 
contains a vulnerability that allows the extension/file name to be tampered with when downloading files.

This is known as a Reflect File Download attack.
https://www.blackhat.com/docs/eu-14/materials/eu-14-Hafif-Reflected-File-Download-A-New-Web-Attack-Vector.pdf

This vulnerability allows a Threat Actor to replace what was `.txt` when uploading with `.sh`, etc. when (victim) downloading.

This is due to improper handling of the filename in the Content-Disposition response header. 
https://github.com/gin-gonic/gin/blob/a889c58de78711cb9b53de6cfcc9272c8518c729/context.go#LL1059-L1059C86

My calculations CVSS (v3):

 * CVSS:3.0/AV:N/AC:H/PR:N/UI:R/S:U/C:H/I:H/A:H
 * Score: 7.5

However, I think the impact of the actual situation is much lower than that of CVSS.

### Env

OS: macOSVentura(13.1)
Go ver: 1.18.2 darwin/arm64
gin ver: v1.9.0  (The latest ver has not been corrected either, so the problem probably remains.)
Browser: Chrome 111.0.5563.110 (arm64)

(Linux or MacOS is required. 
This is because Windows does not allow file names containing " (double-quote) .)

----------------

1. Create File

```
$ touch 'malicious.sh";dummy=.txt'
```

2. Setup gin project

```
$ mkdir myproject
$ go mod init tidy
$ go get -u github.com/gin-gonic/gin
```

3. Generate Vuln code
```
$ vi main.go
```

Edit Directory
```
package main

import (
  "github.com/gin-gonic/gin"
)

func main() {
  r := gin.Default()
  r.GET("/download", func(c *gin.Context) {
    dir := "/Users/{CHANGE_PROJECT_DIRECTRY}/"

    // Although the file name is hard-coded, we assume that the file name is actually determined by the DB or user input.
    filename := "malicious.sh\";dummy=.txt"
    c.FileAttachment(dir + filename, filename)
  })
  r.Run()
}

```

4. Build & Run & Access
```
$ go build ./main.go
$ ./main
```

Access: `http://localhost:8080/download`

------------------

Return Response Header:

[...]

PR #3556, opened 2023-04-01
fix lack of escaping of filename in Content-Disposition

motoyasu-saburi:
https://github.com/gin-gonic/gin/issues/3555

### DIFF
--- a/context.go
+++ b/context.go
@@ -1056,7 +1056,7 @@ func (c *Context) FileFromFS(filepath string, fs http.FileSystem) {
 // On the client side, the file will typically be downloaded with the given filename
 func (c *Context) FileAttachment(filepath, filename string) {
 	if isASCII(filename) {
-		c.Writer.Header().Set("Content-Disposition", `attachment; filename="`+filename+`"`)
+		c.Writer.Header().Set("Content-Disposition", `attachment; filename="`+strings.Replace(filename, "\"", "\\\"", -1)+`"`)
 	} else {
 		c.Writer.Header().Set("Content-Disposition", `attachment; filename*=UTF-8''`+url.QueryEscape(filename))
 	}
--- a/context.go
+++ b/context.go
@@ -1052,11 +1052,17 @@ func (c *Context) FileFromFS(filepath string, fs http.FileSystem) {
 	http.FileServer(fs).ServeHTTP(c.Writer, c.Request)
 }
 
+var quoteEscaper = strings.NewReplacer("\\", "\\\\", `"`, "\\\"")
+
+func escapeQuotes(s string) string {
+	return quoteEscaper.Replace(s)
+}
+
 // FileAttachment writes the specified file into the body stream in an efficient way
 // On the client side, the file will typically be downloaded with the given filename
 func (c *Context) FileAttachment(filepath, filename string) {
 	if isASCII(filename) {
-		c.Writer.Header().Set("Content-Disposition", `attachment; filename="`+strings.Replace(filename, "\"", "\\\"", -1)+`"`)
+		c.Writer.Header().Set("Content-Disposition", `attachment; filename="`+escapeQuotes(filename)+`"`)
 	} else {
 		c.Writer.Header().Set("Content-Disposition", `attachment; filename*=UTF-8''`+url.QueryEscape(filename))
 	}
--- a/context_test.go
+++ b/context_test.go
@@ -1035,8 +1035,8 @@ func TestContextRenderAttachment(t *testing.T) {
 func TestContextRenderAndEscapeAttachment(t *testing.T) {
 	w := httptest.NewRecorder()
 	c, _ := CreateTestContext(w)
-	maliciousFilename := "tampering_field.sh\";dummy=.go"
-	actualEscapedResponseFilename := "tampering_field.sh\\\";dummy=.go"
+	maliciousFilename := "tampering_field.sh\"; \\\"; dummy=.go"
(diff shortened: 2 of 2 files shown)


context.go
```diff
@@ -1056,7 +1056,7 @@ func (c *Context) FileFromFS(filepath string, fs http.FileSystem) {
 // On the client side, the file will typically be downloaded with the given filename
 func (c *Context) FileAttachment(filepath, filename string) {
 	if isASCII(filename) {
-		c.Writer.Header().Set("Content-Disposition", `attachment; filename="`+filename+`"`)
+		c.Writer.Header().Set("Content-Disposition", `attachment; filename="`+strings.Replace(filename, "\"", "\\\"", -1)+`"`)
```
fabricioereche, line 1059: This can fix this specific situation (that enables RFD attack) but what about using [filePath#Clean](https://pkg.go.dev/path/filepath#Clean)?

motoyasu-saburi: I tried using that method, 
but the `"` was not escaped (or URL Encoded), so The vulnerability was not fixed.

fabricioereche: But since it's for file reading those quotes will be needed? File params would be readable with or without it, isn't?

motoyasu-saburi: @fabricioereche 

Perhaps you are confusing the file name with the file path.
There may be some misunderstandings, so I will add to this.

--------------

The first thing to escape is the `filename`. Not `filepath`.

The second thing, The Filename escape requirements are then as follows
 * `"` --> `%2F`, `\"`
 * `\r` --> `%0D`, `\\r`
* `\n` --> `%0A`, `\\n`

The third thing, 
This line is the process of "escaping the 'file name at download time' on the HTTP Response Header.
In other words, filepath is irrelevant.

(The part that actually reads the file is the following line (`filepath`) )
https://github.com/gin-gonic/gin/pull/3556/files#diff-552f47512a00afe5fc6850cc9ddc830a6daeca162750e50aab4ed549685e0253R1063

---------------------

> But since it's for file reading those quotes will be needed? File params would be readable with or without it, isn't?

This is answered by the third matter mentioned above.
This tells the Request side to "download the file with this kind of file name".
It is not the file path to be read by the Server side.

Does this answer your question?

fabricioereche (approved):

zpavlinovic:
Is there a plan for this fix to be merged in?

motoyasu-saburi:
@jerbob92 

> Is there a reason we can't use url.QueryEscape here? Just like in the UTF-8 version?
Both will solve the security issue. 
First, the requirement for encoding by URL Encode is Whtat's WG > multipart/form-data requirements. 

In this case, the requirement to "perform URL Encoding" is not (strictly) the same as that of What's WG, since this is a response.

On the other hand, there is no clear escape requirement specified in the RFC 6266 either.
https://datatracker.ietf.org/doc/html/rfc6266#section-5

After that, we checked about 80 systems of several services, WebFrameworks, Browsers, etc., and found that most of them use BackSlash encoding for response headers, so we chose BackSlash style escaping and implemented it this time.

(For example, Gmail file downloads would have been in backslash format.)

In my personal opinion, 
either is honestly fine, since the root cause is that the Content-Disposition specification is not strict.

jerbob92:
@motoyasu-saburi I tested some different things and couldn't get around this fix, so for the least invasive change, I would indeed go for the string replacer version, and not the `url.QueryEscape`. I would go with the `escapeQuotes` method from `http/multipart` just so that the logic is matched.

motoyasu-saburi:
@jerbob92  Thank you, 
As you said, there was a problem.
I tried to fix it, Could you review it?

jerbob92 (approved):

jerbob92:
@motoyasu-saburi 
I tried the escape trick too, but it didn't work, Chrome just transformed it into an underscore, so filename becomes `tampering_field.sh_`, so that's why I didn't see it as a real problem. But I think the current fix is the best solution.

motoyasu-saburi:
@fabricioereche 
I have made some corrections after receiving your approval, 
so could you please review it again?

hmert (approved):
This fix is very important to us.

motoyasu-saburi:
Could you review PR ?
@thinkerou  @appleboy

zadil (approved):

fabricioereche (approved):

fabricioereche:
> Can we get this in @fabricioereche? Lot's of folks waiting for a patch for [GHSA-2c4m-59x9-fr2g](https://github.com/advisories/GHSA-2c4m-59x9-fr2g)

all set for me

sm003ash:
Hi, when will this be merged and available in the next release? We use the gin framework at my workplace, and this CVE issue is urgent for us to fix, else we will have to migrate to another framework, which we don't want to do.

thinkerou (approved):

thinkerou:
@appleboy please review, thanks!

jclebreton:
This PR is referenced in the Govulcheck database (https://pkg.go.dev/vuln/GO-2023-1737) and many people are waiting for the fix.

AndrewSav:
@appleboy is there anything we can do to help move this forward? It has been almost 2 months with this vulnerability and no fix, people get anxious, feeling unprotected. A suggestion from me would be that if you cannot do the review and merge, could you possibly delegate it to @thinkerou? Many thanks!

hmert:
Unfortunately, we decided to move to another framework. For your security, take action.

thinkerou:
wait #3620 thanks all!

PR #3573, opened 2023-04-21
fix misuses of a vs an

cuishuang:


darth-raijin (changes requested):
The uses are correct

cuishuang:
> The uses are correct

There also seems to be a preference for using an in Go projects.   https://github.com/search?q=repo%3Agolang%2Fgo%20an%20http&type=code

for reference  https:www.a-or-an.com/a_an/http

Issue #2203, opened 2020-01-09
c.ShouldBind()  maybe will stack overflow


vvlgo:
- With issues:
  - Use the search tool before opening a new issue.
  - Please provide source code and commit sha if you found a bug.
  - Review existing issues and provide feedback or react to them.

## Description

I used a structure, and the type of a field is itself. When I upload form data and receive it with shoudbind, the stack overflows. Please use the following code to reproduce. I hope to solve this problem. Thank you
## How to reproduce

```
package main

import (
	"github.com/gin-gonic/gin"
)
//post form-data a
func main() {
	r := gin.Default()
	r.GET("/test", TestShoudbind)
	r.Run(":8080")
}

func TestShoudbind(c *gin.Context){
	a := A{}
	c.ShouldBind(&a)
	c.JSON(200,a)
}

type A struct {
	B string `form:"work_year" json:"b"`
	L *A     `json:"l"`
}
```

## Expectations

```
postman 
post http://localhost:8080/test
form-data
a test111
```
json:{"b":"test111","l":nil}
## Actual result

```
postman 
post http://localhost:8080/test
form-data
a test111

runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow
```

## Environment

- go version:1.3.5
- gin version (or commit ref):1.4.0
- operating system:macos10.15.2

JellyZhang:
Duplicate of #1978 
This problem is caused by circular reference.
You can ignore this field with tag `form:"-"` like this:
```
type A struct {
	B string `form:"b" json:"b"`
	L *A     `form:"-" json:"l"`
}
```

Issue #1978, opened 2019-07-05
Wrong Content-Type leads to stack overflow


k-tipp:
- go version: 1.12.6
- gin version (or commit ref): v1.4.0-dev
- operating system: win10

## Description
Stupid people do stupid things. In my case I selected <application/javascript> as Content-Type in Postman instead of <application/json> and was wondering why the heck my code doesn't work as expected.... well, reason found. Anyway an error message which is a little more specific than the following one would have been nice.

Note: I only tested javascript vs. json header, the error is repeatable and it seems that the content of the json message is not relevant.

## Relevant code snipped
```
func(c *gin.Context) {
     org := Organization{}
    if err := c.ShouldBind(&org); err == nil { // within c.ShouldBind() the stack overflow happens
	// not relevant
	c.Header("Location", loc.String())
    } else {
        // Never reached in this case
        log.Println(err)
    }
}
```

## Error message
```
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow

runtime stack:
runtime.throw(0xd9f020, 0xe)
	c:/go/src/runtime/panic.go:617 +0x79
runtime.newstack()
	c:/go/src/runtime/stack.go:1041 +0x7b7
runtime.morestack()
	c:/go/src/runtime/asm_amd64.s:429 +0x97

goroutine 20 [running]:
github.com/gin-gonic/gin/binding.tryToSetValue(0xc87660, 0xc00307af30, 0x198, 0xc53b26, 0x2, 0x0, 0x0, 0xe7a280, 0xc87660, 0xc53b2a, ...)
	D:/gopath/src/github.com/gin-gonic/gin/binding/form_mapping.go:108 +0x6cc fp=0xc020801458 sp=0xc020801450 pc=0xa5463c
github.com/gin-gonic/gin/binding.mapping(0xc87660, 0xc00307af30, 0x198, 0xc53b26, 0x2, 0x0, 0x0, 0xe7a280, 0xc87660, 0xc53b2a, ...)
	D:/gopath/src/github.com/gin-gonic/gin/binding/form_mapping.go:74 +0x4d1 fp=0xc0208016b0 sp=0xc020801458 pc=0xa53a81
github.com/gin-gonic/gin/binding.mapping(0xcfc580, 0xc00307af30, 0x199, 0xc14a7d, 0x7, 0x0, 0x0, 0xe7a280, 0xcbf9c0, 0x0, ...)
[...]

PR #3586, opened 2023-04-28
refactor: replace a panic with an error when walking too deep on form mapping

thxCode:
this PR is going to report an error rather than panic with stack overflow when walking too deep on form(or customized tag) mapping.

this PR also uses another commit to solve the scenario where multiple binders work on the same struct together.

mitigate some cases like #2203, #1978, and guide users to use `json binder` in a tree-like struct.

### DIFF
--- a/binding/form_mapping.go
+++ b/binding/form_mapping.go
@@ -17,7 +17,8 @@ import (
 )
 
 var (
-	errUnknownType = errors.New("unknown type")
+	errUnknownType      = errors.New("unknown type")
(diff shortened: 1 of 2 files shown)


thxCode:
remove the last commit to keep compatibility with previous versions. https://github.com/gin-gonic/gin/actions/runs/4827373947/jobs/8633084284?pr=3586.

thxCode:
@appleboy  PTAL

PR #3590, opened 2023-04-30
feat: add Error if the request can't be authorized

### LINKED ISSUE #3576: Detect StatusUnauthorized after BasicAuthForRealm middleware has executed
## Description

I would like to add a middleware to Gin which detects whether the HTTP status code has been set by another handler to anything from `400` and above, and if so, add an `application/problem+json` response body that is tailored to the status code.

With the built-in `BasicAuthForRealm` middleware, it currently seems impossible to detect that it has invoked `c.AbortWithStatus(http.StatusUnauthorized)` in my middleware, since `c.Errors()` is not invoked and I can't figure out in my handler that `http.StatusUnauthorized` is being set as the status code.

https://github.com/gin-gonic/gin/blob/a889c58de78711cb9b53de6cfcc9272c8518c729/auth.go#L45-L65

`c.Request.Response.StatusCode` is also unavailable, as `c.Request.Response` is `nil`.

## How to reproduce

```go
package main

import (
	"github.com/gin-gonic/gin"
)

func main() {
	g := gin.Default()

	g.Use(func(c *gin.Context) {
		c.Next()

		if c.IsAborted() {
			c.JSON(200, gin.H{"message": "You are probably not authorized, but I can't tell for sure."})
		}
	})

	accounts := gin.Accounts{"admin": "password"}
	admin := g.Group("admin", gin.BasicAuth(accounts))
	admin.GET("", func(c *gin.Context) {
		c.JSON(200, gin.H{"message": "Hello admin, nice to meet you!"})
	})

	g.Run(":9000")
}
```

## Expectations

I expect to be able to detect what previously run handlers have done to the response, so I can act upon it and enrich the response accordingly.

## Actual result

It is currently not possible to see what a previously run handler have done to the response.

## Environment

- go version: `go1.20.3 darwin/amd64`
- gin version (or commit ref): `v1.9.0`
- operating system: macOS `11.7.6`

asbjornu:
Add `Error` to `gin.Context` if the `BasicAuth` middleware is unable to successfully authorize the request, such that other custom middleware can detect that the request is unauthorized.

A more scalable approach would be to expose the HTTP `status` set by other middleware, but that requires a bigger change than I was comfortable with. I would love to move this in that direction, though, if I'm provided with a bit of guidance.

Resolves #3576.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -7,6 +7,7 @@ package gin
 import (
 	"crypto/subtle"
 	"encoding/base64"
+	"errors"
 	"net/http"
 	"strconv"
 
@@ -25,6 +26,11 @@ type authPair struct {
 }
 
 type authPairs []authPair
+
+var (
+	// ErrUnauthorized cannot authorize the request.
+	ErrUnauthorized = errors.New("unauthorized")
+)
 
 func (a authPairs) searchCredential(authValue string) (string, bool) {
 	if authValue == "" {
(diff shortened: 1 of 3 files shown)


asbjornu:
I just discovered `gin.Context.Writer.Status()` which seems to return `401` as expected. I should make do with it and won't need the functionality this PR provides. Closing.

PR #3604, opened 2023-05-12
change function type to pointer receivers error

aliheydarabadii:
Struct Error has methods on both value and pointer receivers. Such usage is not recommended by the Go Documentation.


aliheydarabadii:
Would anyone be able to review this merge request?

PR #3662, opened 2023-07-09
added proper errors for GetString() in context.go

Danial-Gharib:
- With pull requests:

### DIFF
--- a/context.go
+++ b/context.go
@@ -6,6 +6,7 @@ package gin
 
 import (
 	"errors"
+	"fmt"
 	"io"
 	"log"
 	"math"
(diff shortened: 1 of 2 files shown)


momsspaghettti (commented):
Don't use reflect without necessity please and provide more tests to cover all possible cases

context.go
```diff
@@ -274,13 +275,22 @@ func (c *Context) MustGet(key string) any {
 }
 
 // GetString returns the value associated with the key as a string.
-func (c *Context) GetString(key string) (s string) {
+func (c *Context) GetString(key string) (s string, err error) {
```
flc1125, line 278: This is an adjustment that is not backward compatible, and it is not recommended to make such an adjustment.

flc1125 (changes requested):

xiaotushaoxia:
Your change has caused the GetString function to return two values, breaking backward compatibility, which is unacceptable. Moreover, why are you only focusing on modifying GetString? Don't you also want to modify GetBool and GetInt64?

PR #3687, opened 2023-08-07
Use magenta color for HTTP 101 status in the log

afbjorklund:
When using http.Hijacker, there was a scary looking red log entry.

Color it in a more "neutral" color instead, since it is harmless.

### DIFF
--- a/logger_test.go
+++ b/logger_test.go
@@ -310,6 +310,7 @@ func TestColorForStatus(t *testing.T) {
 		return p.StatusCodeColor()
 	}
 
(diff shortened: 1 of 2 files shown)


logger_test.go
```diff
@@ -310,6 +310,7 @@ func TestColorForStatus(t *testing.T) {
 		return p.StatusCodeColor()
 	}
 
+	assert.Equal(t, magenta, colorForStatus(http.StatusSwitchingProtocols), "101 should be magenta")
```
akshayShirude, line 313: Can we make it for 1xx?

afbjorklund:
It ended up white instead.

* #3741

PR #3669, opened 2023-07-18
docs: changed documentation link for trusted proxies

caleflat:
https://github.com/gin-gonic/gin/pull/3575

Fixes more incorrect links from the linked PR.

PR #3776, opened 2023-11-14
Change URL of Don't Trust All Proxies warning

danielalexis:
This changes the outdated URL to the new Github URL


danielalexis:
Just saw #3669 exists

PR #3904, opened 2024-03-25
feat(binding): add BindPlain

guonaihong:
https://github.com/gin-gonic/gin/pull/2037

```go
package main

import (
	"github.com/gin-gonic/gin"
)

func main() {
	router := gin.Default()
	router.POST("/", func(c *gin.Context) {
		var s string
		c.BindPlain(&s)
		c.String(200, s)
	})

	router.Run()
}

// client
// curl -d "test string" 127.0.0.1:8080
// output
// test string
```

### DIFF
--- a/context.go
+++ b/context.go
@@ -667,6 +667,11 @@ func (c *Context) BindTOML(obj any) error {
 	return c.MustBindWith(obj, binding.TOML)
 }
 
+// BindPlain is a shortcut for c.MustBindWith(obj, binding.Plain).
+func (c *Context) BindPlain(obj interface{}) error {
+	return c.MustBindWith(obj, binding.Plain)
+}
+
 // BindHeader is a shortcut for c.MustBindWith(obj, binding.Header).
 func (c *Context) BindHeader(obj any) error {
 	return c.MustBindWith(obj, binding.Header)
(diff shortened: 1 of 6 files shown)


guonaihong:
It seems that running go test -tags "sonic avx" -v -covermode=count -coverprofile=profile.out will cause a panic. However, I couldn't reproduce this panic because my CPU doesn't support AVX 512 instructions

@appleboy

context.go
```diff
@@ -614,7 +614,7 @@ func (c *Context) SaveUploadedFile(file *multipart.FileHeader, dst string) error
 	}
 	defer src.Close()
 
-	if err = os.MkdirAll(filepath.Dir(dst), 0750); err != nil {
+	if err = os.MkdirAll(filepath.Dir(dst), 0o750); err != nil {
```
RedCrazyGhost, line 617: Caution should be exercised regarding permission modifications to files

guonaihong: These two lines of code are actually the same, except for the representation of file permissions using different formats for octal notation.

In the first line, the file permissions are represented using octal notation, specifically 0750. In octal notation, a number starting with 0 denotes that it's an octal number. Therefore, 0750 represents permissions set to rwxr-x--- (i.e., the owner has read, write, and execute permissions, the group users have read and execute permissions, and others have no permissions).

In the second line, the new octal literal notation is used, denoted by 0o750. In this notation, numbers start with 0o to indicate octal representation. So, 0o750 also represents permissions set to rwxr-x---.

In summary, these two lines of code are entirely equivalent; they just use different formats for representing permissions when using octal notation.

appleboy (approved):

PR #3982, opened 2024-05-29
feat(binding/json): check validity when Bind json

demoManito:
## Description
json was invalid and EOF was returned, so the problem could not be accurately located. so modified `Bind` and `BindBody` to check the legitimacy of json string before decoding.


RedCrazyGhost:
json This part of the module is not suitable for use `unexpected EOF`

Such a function may still need further thinking and discussion.

demoManito:
> The current situation is that an incorrect JSON format returns `unexpected EOF`. After this PR, it will return the error `invalid json`

Hello, my previous description might not have been very clear.

The current situation is that an incorrect JSON format returns `unexpected EOF`. After this PR modification, it will return the error `invalid json`.

demoManito:
@appleboy PTAL

PR #4019, opened 2024-07-28
logger: allow skipping query string output

USA-RedDragon:
This is useful for APIs that might have sensitive information in the query string, such as API keys.

This patch does not change the default behavior of the code unless the new `SkipQueryString` config option is passed in as `true`.

The "skip" term is a bit of a misnomer here, as this doesn't actually skip that log, but modifies the output. I'm open to suggestions for a more appropriate name.


USA-RedDragon:
@austinheap may I ask why this has been closed?

PR #4026, opened 2024-08-07
refine: return error when response writer doesn't plement hijacker

xieyuschen:
The change tries to align with the H2C behavior, which returns an error when the response writer doesn't implement the hijacker api. https://github.com/golang/net/blob/master/http2/h2c/h2c.go#L137 

The panic could be demonstrated as the gist reveals: https://gist.github.com/xieyuschen/ff525642e576de14729b9b9214c656f5

I raised this PR because during our H2C integration with gin, we found this potential panic which is better to be fixed from the framework side. It's an improvement specifically to the HTTP/2 case. It's not a breaking change and could help integrate gin with H2C.


xieyuschen:
Hi @appleboy @manucorporat how do you think about this change?

xieyuschen:
@thinkerou

PR #4029, opened 2024-08-14
feat(context): methods to get nested map from query string

dorzepowski:


dorzepowski:
1. Recently I fixed the issue that makes one test sometimes failing 
2. Also I added support for arrays at the value level of the nested map, while implicit arrays are still returning only first value

In other words: 
```
GET /get?filter[names][]=alice&filter[names][]=bob
```
will produce map
```
map[names:[alice bob]
```

while:
```
GET /get?filter[status]=new&filter[status]=old HTTP/1.1
```

will produce the map with only single value at given key
```
map[status:new]
```

Issue #4074, opened 2024-10-14
reopen "Gin Context implementation breaks context.Context contract #3888"


huangjilaiqin:
this commit did not fix problem,just replace 0 to ContextRequestKey
[#3897](https://github.com/gin-gonic/gin/pull/3897)
![image](https://github.com/user-attachments/assets/497cc3e5-7a70-42ff-9232-0959d6fe6e51)

this code can solve this problem
![image](https://github.com/user-attachments/assets/92be54da-e9d8-4970-8bee-d56502f7db12)

huangjilaiqin:
@FarmerChillax @appleboy check this please

FarmerChillax:
see: https://github.com/gin-gonic/gin/issues/3993#issuecomment-2172593121

FarmerChillax:
> this code can solve this problem

this is a break change, I would not recommend making such changes in a widely used framework.

huangjilaiqin:
this is the root cause, ContextRequestKey==0 make otel's func SpanFromContext can not get the Span but get the c.Request
![image](https://github.com/user-attachments/assets/fde0a985-7cd2-4a62-a4de-c09136cc02cc)
![image](https://github.com/user-attachments/assets/9076888b-1a56-46aa-8d98-d019b15eda44)

FarmerChillax:
In otel the currentSpanKey is Type of traceConrextKeyType.  In here, you can set `ContextWithFallback = true `  to solue this problem

see: https://github.com/gin-gonic/gin/issues/3993#issuecomment-2172593121

with Go Context standards:
```
The provided key must be comparable and should not be of type string or any other built-in type to avoid collisions between packages using context. Users of WithValue should define their own types for keys. To avoid allocating when assigning to an interface{}, context keys often have concrete type struct{}. Alternatively, exported context key variables' static type should be a pointer or interface.
```

ref: https://pkg.go.dev/context#WithValue

FarmerChillax:
Minimum case:
```
package main

import (
	"context"
	"fmt"
	"log"

	"github.com/gin-gonic/gin"
	"go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin"
	"go.opentelemetry.io/contrib/propagators/b3"
	"go.opentelemetry.io/otel"
	"go.opentelemetry.io/otel/propagation"
	"go.opentelemetry.io/otel/sdk/resource"
	sdktrace "go.opentelemetry.io/otel/sdk/trace"
[...]

PR #4075, opened 2024-10-14
fixed Value do not respect the context.WithValue contract

huangjilaiqin:
fixed Value func do not respect the context.WithValue contract [#4074](https://github.com/gin-gonic/gin/issues/4074)


huangjilaiqin:
@appleboy look at this please

FarmerChillax:
This is not the root cause of the current problem, because the types are not consistent.

see: https://go.dev/play/p/3PT-KEnVL1e

PR #4095, opened 2024-11-15
Dev json engine

### LINKED ISSUE #4094: Is it possible to know what render was used inside a middleware function?
## Description

I am creating a middleware function, and I want to modify the json that is sent back after the handler function is run. The thing is that, that json could be created in different ways:

```go
c.IndentedJSON(http.StatusOK, body)
c.JSON(http.StatusOK, body)
...
```

In my middleware function, _I would like to modify that json_, and return it with the same "rendering" it was created, so, as an example, if it was using the **indented json**, I want to modify the json e.g, by wrapping it in another json, but return it in indented json format too, as it was originally written in the handler function.

## How to reproduce

```go
package main

import (
	"github.com/gin-gonic/gin"
)

func myMiddleware(c *gin.Context) {
   // do something here with the json response
}

func main() {
	g := gin.Default()
        g.Use(myMiddleware)
	g.GET("/hello/:name", func(c *gin.Context) {
                body := map[string]string{"name": "Manuel", "surname":"gin"}
		c.IndentedJSON(200, body)
	})
	g.Run(":9000")
}
```

## Expectations

So in case that the middleware wraps the response, I would expect something like:
```
$ curl http://localhost:9000/hello/world
{
  "data": {
    "name": "Manuel",
    "surname": "gin"
  }
}
```

## Actual result

```
$ curl -i http://localhost:9000/hello/world
{"data": {"name":"Manuel","surname": "gin"}}
```
Because I don't know how it was "rendered", I assumed it was using the normal c.JSON

## Environment

- go version: 1.23
- gin version (or commit ref):
- operating system: windows

## Comment

Would it make sense to add a key in the context specifying what render was used?
Something like:
```
c.Set("gin:render", "indentedJSON")
```

ChenPuChu:
- Implemented a user-customizable JSON render function
  - This function does not disrupt the implementation of the original function
  - This function is implemented through the factory mode
  - This is my first time submitting a fix. Maybe my code is not standard. If there is anything wrong, please feel free to correct me.
 
Fixes #4094

### DIFF
--- a/context.go
+++ b/context.go
@@ -87,7 +87,7 @@ type Context struct {
 	// or PUT body parameters.
 	formCache url.Values
 
-	// SameSite allows a server to define a cookie attribute making it impossible for
+	// sameSite allows a server to define a cookie attribute making it impossible for
 	// the browser to send this cookie along with cross-site requests.
 	sameSite http.SameSite
 }
--- a/context.go
+++ b/context.go
@@ -52,6 +52,26 @@ const ContextRequestKey ContextKeyType = 0
 // abortIndex represents a typical value used in abort functions.
 const abortIndex int8 = math.MaxInt8 >> 1
 
+// defaultJsonRenderName, default JSON render name
+const defaultJsonRenderName string = "render.JSON"
+
+// jsonRenderRegistry, storing JSON render types and their corresponding creation factory functions
+var jsonRenderRegistry = make(map[string]func(data any) render.Render)
+
+// RegisterJsonRender allows users to register JSON render factory functions
+func (c *Context) RegisterJsonRender(name string, jsonRender func(data any) render.Render) {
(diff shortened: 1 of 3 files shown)


ChenPuChu:
@JimChenWYU

JimChenWYU:
I don't know your pr how to fix his problem...

In my understood， he just want to change response body in his gin-middleware

manuelarte:
@JimChenWYU thanks for answering. To be more clear, what I want is that, when modifying the response, I want to keep the same "format" the user used.

So, my middleware can't know whether the user used `c.indentedjson`, or `c.json` to format the output

ChenPuChu:
@manuelarte Okay, this is your guideline

manuelarte:
So, in my middleware function, what would I need to write to retrieve what render was used?

Issue #4120, opened 2024-12-23
[Feature]: `ShouldBindJSON` return `EOF` when the request body is an empty string.


tttoad:
- With issues:
  - Use the search tool before opening a new issue.
  - Please provide source code and commit sha if you found a bug.
  - Review existing issues and provide feedback or react to them.

## Description

`ShouldBindJSON` return `EOF` when the request body is an empty string.
This error message is not friendly.

## How to reproduce

```
package main

import (
	"fmt"

	"github.com/gin-gonic/gin"
)

type R struct {
	C string `form:"c" binding:"required"`
}

func main() {
	g := gin.Default()
	g.POST("/hello", func(c *gin.Context) {
		if err := c.BindJSON(&R{}); err != nil {
			fmt.Println(err, len(c.Errors))
		}
	})
	g.Run(":9000")
}
```

## Expectations

```
$ curl --request POST '127.0.0.1:9000/hello' --data '' 
```

## Actual result

```
print EOF 1.
Expect: 
`Key: 'R.C' Error:Field validation for 'C' failed on the 'required' tag 0` or `invalid request` or `empty body`


```

## Environment

- go version:
- gin version (or commit ref):
- operating system:

PR #4122, opened 2024-12-26
a temporary fix for issue 4120

alsammak-array:
According to the issue #4120 

a workaround is to read the body and reset it to {} inside `BindJSON(args)`
```go 
// BindJSON is a shortcut for c.MustBindWith(obj, binding.JSON).
func (c *Context) BindJSON(obj any) error {

	body, err := io.ReadAll(c.Request.Body)
	if err != nil {
		return fmt.Errorf("failed to read body: %v", err)
	}

	// If the body is empty, set it to {}
	if len(body) == 0 {
		body = []byte("{}")
		c.Request.Body = io.NopCloser(bytes.NewBuffer(body)) // Set to {} to trigger the error
		return c.MustBindWith(obj, binding.JSON)
	}

	// Restore the body since it was read
	c.Request.Body = io.NopCloser(bytes.NewBuffer(body))
	return c.MustBindWith(obj, binding.JSON)
}
```
the output should be as expected in the issue.

### DIFF
--- a/context.go
+++ b/context.go
@@ -5,7 +5,9 @@
 package gin
 
 import (
+	"bytes"
 	"errors"
+	"fmt"
 	"io"
 	"io/fs"
 	"log"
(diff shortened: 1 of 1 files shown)


Ale-Cas (commented):
Thanks for this and LGTM, I would possibly just add a test case in https://github.com/gin-gonic/gin/blob/master/context_test.go#L1866 for this behavior.

IMO those requests should not error out and would prefer this solution over https://github.com/gin-gonic/gin/pull/4124

realpg:
LGTM.
Hope to merge this soon.

PR #4166, opened 2025-02-26
fix: correct RLock scope in Context.Copy method

1911860538:


NezhaFan:
You are wrong,  must add a read lock during the loop to prevent writing at this time.   and maybe you don't understand the copy of map .

1911860538:
> You are wrong, must add a read lock during the loop to prevent writing at this time. and maybe you don't understand the copy of map .

`c.mu.RLock()` is used to protect `c.Keys`, not `cp.Keys`. `cp.Keys` does not require lock protection here because the `Copy` method is in the process of creating `cp`, so there is no read-write race condition for `cp.Keys`.Since the `Copy` method involves reading operations on `c.Keys`, my modification poses no issues.

NezhaFan:
我觉得你在理解map拷贝和读写锁这里有点问题。 
`	var mu sync.RWMutex
	n := 50000

	m1 := make(map[int]int, n)
	for i := 0; i < n; i++ {
		m1[i] = i
	}

	// 并发写入
	for i := 0; i < n; i++ {
		go func(i int) {
			mu.Lock()
			defer mu.Unlock()
			m1[i] = i
		}(i)
	}

	mu.RLock()
	m2 := m1
	// 你说拷贝后就解锁
	mu.RUnlock()

	// 拷贝
	cp := make(map[int]int, n)
	for k, v := range m2 {
		cp[k] = v
	}
	// 应该读完再解锁
	// mu.RUnlock()`

PR #4207, opened 2025-03-31
docs: Fixing English grammar mistakes and awkward sentence structure in doc/doc.md

boyczuk:
This PR is editing the English grammar and spelling of several comments and instructions throughout the `docs/doc.md` file. 

While reading the [doc](https://github.com/boyczuk/gin/blob/master/docs/doc.md) for help with authentication I noticed that there are quite a few spots where the instructions had redundant or incorrect grammar. I scanned through and hopefully corrected all of the big ones.

### Example change

**Before:**
```
Skip validate: when running the above example using the above the `curl` command, it returns error.
Because the example uses `binding:"required"` for `Password`.
If you use `binding:"-"` for `Password`, then it will not return error when running the above example again.
```

**After:**
```
Skip-validation: Running the example above using the `curl` command returns an error.
This is because the example uses `binding:"required"` for `Password`.
If instead, you use `binding:"-"` for `Password`, then it will not return an error when you run the example again.
```

### DIFF
--- a/docs/doc.md
+++ b/docs/doc.md
@@ -70,7 +70,7 @@
 
 ### Build with json replacement
 
-Gin uses `encoding/json` as default json package but you can change it by build from other tags.
+Gin uses `encoding/json` as the default JSON package but you can change it by build from other tags.
 
 [jsoniter](https://github.com/json-iterator/go)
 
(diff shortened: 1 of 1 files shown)


docs/doc.md
```diff
@@ -70,7 +70,7 @@
 
 ### Build with json replacement
 
-Gin uses `encoding/json` as default json package but you can change it by build from other tags.
+Gin uses `encoding/json` as the default JSON package but you can change it by build from other tags.
```
som-snytt, line 73: building

boyczuk: Oh crap I completely missed that lol, thanks!

appleboy (approved):

PR #4275, opened 2025-06-23
feat: add new function to get engine from Context type

### LINKED ISSUE #4267: How to get engine from *gin.Context?
Hi,

  Is there a way to get engine from the Context? I have some middleware where I
want to call HandleContext to do a redirect.

-Tomas

yosp313:
This change resolves #4267 by adding option to get engine from context.

Signed-off-by: Youssef Ayman <youssef.ayman992003@gmail.com>

### DIFF
--- a/context.go
+++ b/context.go
@@ -93,6 +93,14 @@ type Context struct {
 	sameSite http.SameSite
 }
 
+// Get Engine from Context to handle contexts
+func (c *Context) GetEngine() *Engine {
+	if c.engine == nil {
+		panic("context engine is nil")
+	}
+	return c.engine
+}
+
 /************************************/
 /********** CONTEXT CREATION ********/
 /************************************/


FarmerChillax:
I think this demand is unreasonable. engine and ctx should not be mixed

yosp313:
> I think this demand is unreasonable. engine and ctx should not be mixed

Sorry I'm just a new developer and i wanted to contribute to this project because i find it the most convenient router framework in golang so can you explain to me why they shouldn't be mixed?

FarmerChillax:
> > I think this demand is unreasonable. engine and ctx should not be mixed
> 
> Sorry I'm just a new developer and i wanted to contribute to this project because i find it the most convenient router framework in golang so can you explain to me why they shouldn't be mixed?

ctx refers to the request context, engine is the application instance, I think the two should not be mixed

yosp313:
> > > I think this demand is unreasonable. engine and ctx should not be mixed
> > 
> > 
> > Sorry I'm just a new developer and i wanted to contribute to this project because i find it the most convenient router framework in golang so can you explain to me why they shouldn't be mixed?
> 
> ctx refers to the request context, engine is the application instance, I think the two should not be mixed

So you mean that it's not good to control the app instance inside of handler functions is that correct?

FarmerChillax:
> > > > I think this demand is unreasonable. engine and ctx should not be mixed
> > > 
> > > 
> > > Sorry I'm just a new developer and i wanted to contribute to this project because i find it the most convenient router framework in golang so can you explain to me why they shouldn't be mixed?
> > 
> > 
> > ctx refers to the request context, engine is the application instance, I think the two should not be mixed
> 
> So you mean that it's not good to control the app instance inside of handler functions is that correct?

Not absolute, just an expression in current context, I think this PR solution is less appropriate than the global variable solution.

Doing so will bring many potential problems, as well as problems with the surrounding ecosystem. For example, concurrency issues

yosp313:
Oh okay i got it now thanks for explaining. I will close this pull request now.

PR #4279, opened 2025-06-27
clean up the multipart request files- this resolves #4278

### LINKED ISSUE #4278: Why no call to Request.MultipartForm.RemoveAll
I found if upload large file, a file in /tmp will keep forevery until next booting.
But go net/server package will call RemoveAll when request finished(https://github.com/golang/go/blob/b5d555991ab73e06e09741952a66dd7eeaf2a185/src/net/http/server.go#L1700)
```go
	if w.req.MultipartForm != nil {
		w.req.MultipartForm.RemoveAll()
	}
```
My request and response seem work, but /tmp/multipart-xxxx file not delete by net/server. If i mannual call RemoveAll, it work ok.

iamritikbhardwaj:
There was this issue that the multipart form is not cleaned after req completion 

```go
// CleanUp clean up the multipart request
func (r *multipartRequest) CleanUp() error {
	if r.MultipartForm != nil {
		return r.MultipartForm.RemoveAll()
	}
	return nil
}
```
this fixes #4278

### DIFF
--- a/binding/multipart_form_mapping.go
+++ b/binding/multipart_form_mapping.go
@@ -32,6 +32,14 @@ func (r *multipartRequest) TrySet(value reflect.Value, field reflect.StructField
 	return setByForm(value, field, r.MultipartForm.Value, key, opt)
 }
 
+// CleanUp clean up the multipart request
+func (r *multipartRequest) CleanUp() error {
+	if r.MultipartForm != nil {
+		return r.MultipartForm.RemoveAll()
+	}
+	return nil
+}
+
 func setByMultipartFormFile(value reflect.Value, field reflect.StructField, files []*multipart.FileHeader) (isSet bool, err error) {
 	switch value.Kind() {
 	case reflect.Ptr:


FarmerChillax:
I don't think this is the optimal solution

iamritikbhardwaj:
> I don't think this is the optimal solution

Any other suggestions 🤔

FarmerChillax:
> > I don't think this is the optimal solution
> 
> Any other suggestions 🤔

The current usage scenario does not seem clear. I think we should first understand the scenario and then design a plan

iamritikbhardwaj:
> > > I don't think this is the optimal solution
> > 
> > 
> > Any other suggestions 🤔
> 
> The current usage scenario does not seem clear. I think we should first understand the scenario and then design a plan

Okay please let me know if a discussion is required as i am a fresher looking to learn things from experienced people.

FarmerChillax:
> > > > I don't think this is the optimal solution
> > > 
> > > 
> > > Any other suggestions 🤔
> > 
> > 
> > The current usage scenario does not seem clear. I think we should first understand the scenario and then design a plan
> 
> Okay please let me know if a discussion is required as i am a fresher looking to learn things from experienced people.

There are related discussions in the issue, welcome to participate and put forward your own opinions

PR #4295, opened 2025-07-13
Prevention of Hijack() runtime panics

suwakei:
## Proposal

The current implementation of the Hijack() method allows it to be called even after the response body has already been written to the client;
Go's internal implementation of http.Server is designed so that calling `Hijack ()` after `Write()` or `WriteHeader()` has been called is designed to return an error,
and the current Gin implementation differs from this.

Because of this difference,
some implementations of the underlying http.ResponseWriter may experience unexpected panics if Hijack() is called at an unintended time.

Therefore, we suggest changing the Hijack() method to check at the beginning of the method to see if the response has already been written
 and return a clear and descriptive error if it has.

### Definition of `errHijackAlreadyWritten`:
- Define an explicit error variable to return when `Hijack()` fails.

### Update `Hijack()` method:
- At the beginning of the method, call `w.Written()` to see if the response has already been written.
- If it has already been written, return an `errHijackAlreadyWritten` error and stop further execution.

### DIFF
--- a/response_writer.go
+++ b/response_writer.go
@@ -6,6 +6,7 @@ package gin
 
 import (
 	"bufio"
+	"errors"
 	"io"
 	"net"
 	"net/http"
@@ -15,6 +16,8 @@ const (
 	noWritten     = -1
 	defaultStatus = http.StatusOK
 )
+
+var errHijackAlreadyWritten = errors.New("gin: response already written")
 
 // ResponseWriter ...
 type ResponseWriter interface {
@@ -106,6 +109,9 @@ func (w *responseWriter) Written() bool {
 
 // Hijack implements the http.Hijacker interface.
 func (w *responseWriter) Hijack() (net.Conn, *bufio.ReadWriter, error) {
+	if w.Written() {
+		return nil, nil, errHijackAlreadyWritten
+	}
 	if w.size < 0 {
 		w.size = 0
 	}
--- a/response_writer_test.go
+++ b/response_writer_test.go
@@ -131,39 +131,54 @@ type mockHijacker struct {
 	hijacked bool
 }
 
+// Hijack implements the http.Hijacker interface. It just records that it was called.
 func (m *mockHijacker) Hijack() (net.Conn, *bufio.ReadWriter, error) {
 	m.hijacked = true
 	return nil, nil, nil
 }
 
 func TestResponseWriterHijackAfterWrite(t *testing.T) {
-	// Test case 1: Hijack before writing
-	hijacker := &mockHijacker{ResponseRecorder: httptest.NewRecorder()}
-	writer := &responseWriter{}
-	writer.reset(hijacker)
-	w := ResponseWriter(writer)
+	tests := []struct {
+		name          string
+		action        func(w ResponseWriter) error // Action to perform before hijacking
+		expectHijack  bool
+		expectWritten bool
+		expectError   error
+	}{
+		{
+			name:          "hijack before write should succeed",
+			action:        func(w ResponseWriter) error { return nil },
+			expectHijack:  true,
+			expectWritten: true, // Hijack itself marks the writer as written
+			expectError:   nil,
+		},
+		{
+			name: "hijack after write should fail",
+			action: func(w ResponseWriter) error {
+				_, err := w.Write([]byte("test"))
+				return err
+			},
+			expectHijack:  false,
+			expectWritten: true,
+			expectError:   errHijackAlreadyWritten,
+		},
+	}
 
-	_, _, err := w.Hijack()
-	require.NoError(t, err)
-	assert.True(t, hijacker.hijacked, "Hijack should be called")
-	assert.True(t, w.Written(), "Written() should be true after Hijack")
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			hijacker := &mockHijacker{ResponseRecorder: httptest.NewRecorder()}
+			writer := &responseWriter{}
+			writer.reset(hijacker)
+			w := ResponseWriter(writer)
 
-	// Test case 2: Hijack after writing
-	hijacker2 := &mockHijacker{ResponseRecorder: httptest.NewRecorder()}
-	writer2 := &responseWriter{}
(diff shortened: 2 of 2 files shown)


suwakei:
All lines introduced in this PR are covered by tests.

The reported -0.29% decrease in coverage is not caused by these changes,
but rather due to a difference in the base reference (commit 3dc1cd6).
If there's anything you'd like me to improve or test further, feel free to let me know!

suwakei:
Hi! Just following up to see if there's anything I can improve in this PR 🙇

OrangeDoro (commented):
Hi! I'm a grad student working on a research project about using large language models to automate code review. Based on your commit be39a2b9bee66dfe1c7405285e6b38ed351faae6 and the changes in response_writer_test.go, my tool generated this comment:
1. **Memory Management**: Ensure that the `ResponseRecorder` in `mockHijacker` is properly managed and not leaked. If `mockHijacker` instances are not cleaned up after use, it could lead to memory leaks.
2. **Concurrency Considerations**: If `mockHijacker` is used in a concurrent context, consider using synchronization mechanisms (like mutexes) to protect access to shared state.
3. **Error Handling in `Hijack` Method**: In the `Hijack` method of `mockHijacker`, consider returning meaningful values instead of `nil, nil, nil`, or at least a comment indicating that this is a mock implementation.
4. **Testing Edge Cases**: The test cases provided do not cover scenarios where the `Write` method might fail. Add tests that simulate such failures to ensure that the `Hijack` method behaves correctly under those conditions.
5. **Remove Duplicate Type Definition**: The `mockHijacker` type is defined twice in the patch. This is unnecessary and can lead to confusion. Remove the duplicate definition.
6. **Group Related Assertions**: Group related assertions together and add comments to clarify the purpose of each assertion for better readability.
7. **Assertion Messages**: Ensure that the error messages in assertions are clear and consistent in style. For example, rephrase "Hijack after write should return errHijackAlreadyWritten" to match other messages.
8. **Consider Logging**: In the `Hijack` method, consider adding logging to trace when a hijack occurs, and in the test function, log the state before and after hijack attempts.
9. **Enhance Test Case Comments**: In `TestResponseWriterHijackAfterWrite`, enhance comments for test cases to clarify what each test is verifying.
10. **Parameterize Test Cases**: Consider using table-driven tests to allow for easier addition of new scenarios without duplicating code.
11. **Documentation**: Ensure that the new test cases are documented, explaining the purpose of each test and the expected behavior.

As part of my research, I'm trying to understand how useful these comments are in real-world development. If you have a moment, I'd be super grateful if you could quickly reply to these two yes/no questions:


1. Does this comment provide suggestions from a dimension you hadn’t considered?

2. Do you find this comment helpful?

Thanks a lot for your time and feedback! And sorry again if this message is a bother.

suwakei:
Thank you for your review, I have corrected the points you pointed out.

## 1. Memory Management:
Opinion: This point is correct as a general reminder, but it is not a problem in this code. mockHijacker is created as a local variable in the TestResponseWriterHijackAfterWrite function.
I believe that there is no need to worry about memory leaks because the memory is automatically released at the end of the function.

## 2. Concurrency Considerations:
Opinion: In the current test code, mockHijacker is not used in concurrency, 
I don't think it needs to be supported at the moment.

## 3. Error Handling in Hijack Method:
Opinion: mockHijacker is a mock for testing, and since the return value is not used in the current test, an implementation that returns nil is fine.
However, to increase code readability, I will specify in the comments that this is an intentional implementation.
```go
	hijacked bool
}
// response_writer_test.go
++ // Hijack implements the http.Hijacker interface. It just records that it was called.
func (m *mockHijacker) Hijack() (net.Conn, *bufio.ReadWriter, error) {
	m.hijacked = true
	return nil, nil, nil
```

## 4. Testing Edge Cases:
Opinion:  It is important for the robustness of the code to test that Hijack works as expected even if responseWriter.Write fails. Looking at the implementation of responseWriter.Write,
first WriteHeaderNow() is called and w.size is updated from noWritten(-1) to 0. This causes Written() to be true. This process takes place before the actual w.ResponseWriter.Write(data) succeeds or fails.
Thus, even if Write fails, subsequent Hijack() should return errHijackAlreadyWritten because Written() is already true. Add a test case to guarantee this behavior.

## 5. Remove Duplicate Type Definition:
Opinion:Looking at the code in response_writer_test.go, there is only one definition of mockHijacker.
This is most likely a misrecognition by the AI tool when parsing the difference (PATCH). It is safe to ignore this point.

## 6. Group Related Assertions:
Describe the text in place of documentation in the name field of the test.
(See also 11. Documentation:)

## 7. Assertion Messages:
Fixed style consistency.

## 8. Consider Logging: 
Adding logging to a simple mock like this one may be excessive.

## 9. Enhance Test Case Comments:
Describe the text in place of documentation in the name field of the test.
(See also 11. Documentation:)

## 10. Parameterize Test Cases:
Fixed.
It will be easier to add new test cases like those pointed out in “4. Testing of Edge Cases” in the future It will be easier to add new test cases like those pointed out in "4.

## 11. Documentation: 
 This is related to “9. Enhanced test case comments” and “10. Parameterization of test cases”. We believe that the name field (e.g., “hijack before write should succeed”) that is attached to each entry in a table-driven test serves as documentation as it is.

suwakei:
Hi maintainers 👋 
Just following up on this PR — happy to make any changes if needed. Please let me know if there's anything blocking review. Thanks!

suwakei:
Thanks for the feedback! I've updated the tests to:

- Verify the `Written()` state before and after `Hijack()` as suggested
- Ensure that test expectations align with the actual behavior

Let me know if there's anything else I should update. 🙌

nanmu42:
# Friendly reminder for users of [github.com/coder/websocket](https://github.com/coder/websocket)

[Gin 1.11.0 does not work with github.com/coder/websocket](https://github.com/gin-gonic/gin/issues/4372), use v1.10.1 instead and wait for a new Gin release.

See also:

* https://github.com/coder/websocket/issues/538
* https://github.com/gin-gonic/gin/pull/4373
* https://github.com/coder/websocket/issues/166

PR #4299, opened 2025-07-16
Fixed version number discrepancy

suwakei:
### Problem

There is a mismatch between the Go version check logic and the warning message shown to users:

- Constant: `const ginSupportMinGoVer = 21`
- Message: `"[WARNING] Now Gin requires Go 1.23+\n"`

As a result, users running Go 1.21 or 1.22 cannot see the warning, even though the message states that 1.23+ is required.
## Cause
The logic in `debugPrintWARNINGDefault()`:

```go
func debugPrintWARNINGDefault() {
	if v, e := getMinVer(runtime.Version()); e == nil && v < ginSupportMinGoVer {
		debugPrint(`[WARNING] Now Gin requires Go 1.23+.

`)
```
With ginSupportMinGoVer = 21, however, users on Go 1.21 (v = 21) or 1.22 (v = 22) don't trigger the warning (v < 21 is false).
## Fix
Update the constant ginSupportMinGoVer to 23 to match the warning message indicating that Go 1.23+ is required.

This ensures users running Go versions below 1.23 will properly receive the warning,
aligning behavior with messaging.

### DIFF
--- a/debug.go
+++ b/debug.go
@@ -13,7 +13,7 @@ import (
 	"sync/atomic"
 )
 
-const ginSupportMinGoVer = 21
+const ginSupportMinGoVer = 23
 
 // IsDebugging returns true if the framework is running in debug mode.
 // Use SetMode(gin.ReleaseMode) to disable debug mode.


suwakei:
We have verified that this change has written tests for all modified code.

The Codecov report also shows that all modified lines are covered.

The -0.30% decrease in overall coverage is likely due to the fact that the base branch is 142 commits older than master and does not affect this fix.

We appreciate your confirmation.

appleboy (approved):

PR #4306, opened 2025-07-21
feat: add an option to allow trailing slash insensitive matching

### LINKED ISSUE #915: Support for optional trailing slash
Hi, I tried looking around existing issues but I could not see anything related. So, I would like to know if there is a way to support optional trailing slash at the end of a route.

For example:

`r.POST("/users/?", ...)`

Right now I have to use two posts:

```
r.POST("/users", ...)
r.POST("/users/", ...)
```

charlie0129:
Add an option `TrailingSlashInsensitivity` in engine to allow trailing slash insensitive matching, e.g., if a handler for `/path` is defined, `/path` and `/path/` will go the the same handler without redirection responses.

See the tests for details.

`TrailingSlashInsensitivity` is disabled by default so existing behavior should not be affected.

fixes #915

```
go test -benchmem -run=^$ -bench ^BenchmarkRouteTrailingSlashInsensitivity$ github.com/gin-gonic/gin

goos: darwin
goarch: arm64
pkg: github.com/gin-gonic/gin
cpu: Apple M2
BenchmarkRouteTrailingSlashInsensitivity/Insensitive_Match-8         	  546610	      2095 ns/op	    6150 B/op	      18 allocs/op
BenchmarkRouteTrailingSlashInsensitivity/Sensitive_Match-8           	  589981	      2201 ns/op	    6151 B/op	      18 allocs/op
BenchmarkRouteTrailingSlashInsensitivity/Redirect_Match-8            	  455587	      2776 ns/op	    6474 B/op	      27 allocs/op
PASS
ok  	github.com/gin-gonic/gin	3.805s
```


ddevsr (changes requested):
```suggestion
	if length := len(p); length > 1 && p[length-1] == '/' {
		return p[:length-1];
	}
```

Keep this logic, faster than using `strings.HasSuffix()`.

charlie0129:
@ddevsr Hi, I kept the original logic

gin.go
```diff
@@ -745,6 +769,14 @@ func serveError(c *Context, code int, defaultMessage []byte) {
 	c.writermem.WriteHeaderNow()
 }
 
+func addOrRemoveTrailingSlash(p string) string {
+	ret := p + "/"
+	if length := len(p); length > 1 && p[length-1] == '/' {
+		ret = p[:length-1]
+	}
+	return ret
```
ddevsr:
```suggestion
	if length := len(p); length > 1 && p[length-1] == '/' {
		return p[:length-1]
	}
	return p + "/"
```

No need variable. just return like before your changes

charlie0129: Good point.

Issue #4312, opened 2025-07-23
Missing support for slices with BindUnmarshaler elements


asmeikal:
Currently there is no support for slices that have elements implementing the `BindUnmarshaler` interface, but there is support for slices that _implement directly_ `BindUnmarshaler` (thanks to #3970 and #3971).

This is caused  by `setArray` [always calling `setWithProperType` on its elements](https://github.com/asmeikal/gin/blob/master/binding/form_mapping.go#L450).

This is a test reproducing the problem:

```go
package test

import (
	"errors"
	"github.com/gin-gonic/gin/binding"
	"github.com/stretchr/testify/assert"
	"net/url"
	"strings"
	"testing"
)

type customUnmarshalParamType struct {
	Protocol string
	Path     string
	Name     string
}

func (f *customUnmarshalParamType) UnmarshalParam(param string) error {
	parts := strings.Split(param, ":")
	if len(parts) != 3 {
		return errors.New("invalid format")
	}
	f.Protocol = parts[0]
	f.Path = parts[1]
	f.Name = parts[2]
	return nil
}

func TestSliceOfBindUnmarshal(t *testing.T) {
	var s, expected struct {
		FileData []customUnmarshalParamType `form:"data"`
	}
	form := url.Values{"data": {`file:/foo:happiness`, `http:/bar:sadness`}}
	err := binding.MapFormWithTag(&s, form, "form")
	assert.Nil(t, err)
	expected.FileData = []customUnmarshalParamType{
		{
			Protocol: "file",
			Path:     "/foo",
			Name:     "happiness",
		},
		{
			Protocol: "http",
			Path:     "/bar",
			Name:     "sadness",
		},
	}
	assert.Equal(t, expected, s)
}
```

In this case the mapping fails with a `json.SyntaxError` with a value of `invalid character 'i' in literal false (expecting 'a')`, since the default behaviour for struct fields is to [parse them as JSON](https://github.com/asmeikal/gin/blob/master/binding/form_mapping.go#L336).

This happens for both the latest tagged version (v1.10.1) and using the latest commit from master.

PR #4313, opened 2025-07-23
feat(form): support BindUnmarshaler slice elements (#4312)

asmeikal:
Allows for slice with elements that implement BindUnmarshaler, solving #4312.


takanuva15:
Hey just a heads-up, I think PR #4203 might fix this issue and add other great functionality as well (namely, the ability to use native Golang `encoding.TextUnmarshaler` instead of having to define gin's Unmarshal funciton)

PR #4352, opened 2025-09-14
refactor: use maps.Copy for cleaner map handling

russcoss:
There is a [new function](https://pkg.go.dev/maps@go1.21.1#Copy) added in the go1.21 standard library, which can make the code more concise and easy to read.


russcoss:
@appleboy Hi, I have resolved the conflicts and rebased with the latest code. Please review it when you have time. Thanks.

PR #4371, opened 2025-09-23
perf: optimize route lookup from O(n) to O(1) using hash map for HTTP method resolution

PankajRedekar:
route optimization


PankajRedekar:
not make sense to add hash overhead for just 9 methods max

PR #4422, opened 2025-11-06
fix(gin): close os.File in RunFd to prevent resource leak

1911860538:
`os.NewFile` creates a `*os.File` object that holds a reference to
the file descriptor. Even though the file descriptor itself is
not opened by `os.NewFile`, the `*os.File` object needs to be cleaned
up to release its reference.

### DIFF
--- a/gin.go
+++ b/gin.go
@@ -593,6 +593,7 @@ func (engine *Engine) RunFd(fd int) (err error) {
 	}
 
 	f := os.NewFile(uintptr(fd), fmt.Sprintf("fd@%d", fd))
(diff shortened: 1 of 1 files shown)


1911860538:
Verified safe since net.FileListener internally duplicates the
file descriptor via dup(), creating a separate socket handle.
Closing the original *os.File only releases its own reference and
does not close the underlying network socket used by the listener.

Therefore, defer f.Close() does not interfere with existing connections
or cause "use of closed network connection" errors, including in
graceful restart scenarios where the listener is inherited across processes.

Issue #4451, opened 2025-11-23
router.StaticFS calls fs with invalid file paths


NicoClack:
### Description

Hi. I encountered some issues trying to serve nested embedded HTML files like `tutorials/making-gin/index.html` with [gin-contrib/static](https://github.com/gin-contrib/static) and after some digging, I think the **root cause is an issue in Gin itself**.

The problem is that when you try to visit `example.com/tutorials/making-gin/` (if there's no trailing slash, you get redirected), `http.FS` (which converts the newer `fs.FS` to the older `http.FileSystem` that Gin uses) will correctly remove the prefixed slash before calling the `fs.FS`. However, it's still called with invalid paths like `tutorials/making-gin/`. Many `fs.FS` implementations will reject the trailing slash like `fs.Sub`, which is used by gin-contrib/static, because `fs.ValidPath` considers it to be invalid.

Go's standard library also has a [reference implementation](https://cs.opensource.google/go/go/+/refs/tags/go1.25.4:src/net/http/fs.go;l=988) of a handler which uses `http.FileSystem`, which is `http.FileServer`. It calls `path.Clean` before ultimately calling `fs.Open` and given that Gin's `createStaticHandler` uses `http.FileServer` once it's checked if the file exists, I think the handler should be responsible for cleaning the path before calling an `http.FileSystem`.

Also, the reason why `router.Static` doesn't have this issue is because it uses `http.Dir` (via a wrapper) which seems to be more leniant than a lot of implementations. It calls `path.Clean` before calling `os.Open`. I think it's still a mistake in `router.Static` that it doesn't clean the paths itself though.

PR: https://github.com/gin-gonic/gin/pull/4452

Thanks in advance.

### Gin Version

v1.11.0

### Can you reproduce the bug?

Yes

### Source Code

My PR has what would be a failing test without the fix. Without the fix, there's a 404.

```go
//go:embed testdata/embed
var embeddedFolder embed.FS

const embeddedPath = "testdata/embed"

func TestRouteStaticFSCleansPath(t *testing.T) {
	router := New()
	subFS, err := fs.Sub(embeddedFolder, embeddedPath)
	require.NoError(t, err)
	fs := &OnlyFilesFS{
		FileSystem: http.FS(subFS),
	}
	router.StaticFS("/", fs)
	router.NoRoute(func(c *Context) {
		c.String(http.StatusNotFound, "non existent")
	})

[...]

PR #4452, opened 2025-11-23
Clean paths before calling the `http.FileSystem` passed to `router.StaticFS`

NicoClack:
Fixes the issues I describe in https://github.com/gin-gonic/gin/issues/4451

Some questions:
- Is this considered a breaking change? It changes some behaviour but mostly fixes it
- Are the subtleties of this correct? I adapted [`http.Dir`](https://cs.opensource.google/go/go/+/master:src/net/http/fs.go;l=77?q=http.dir&ss=go%2Fgo) but maybe we want to implement the slightly simpler fix in [`http.FileServer`](https://cs.opensource.google/go/go/+/refs/tags/go1.25.4:src/net/http/fs.go;l=988) instead?
- Should the test use embedding? I'm copying [gin-contrib/static](https://github.com/gin-contrib/static) a bit, but that makes it a bit more realistic?

# Pull Request Checklist

- [ ✔️ ] Open your pull request against the `master` branch.
- [ ❌ ] All tests pass in available continuous integration systems (e.g., GitHub Actions).
- - I think there might be some issues with the CI?
- [ ✔️ ] Tests are added or modified as needed to cover code changes.
- [ N/A ] If the pull request introduces a new feature, the feature is documented in the `docs/doc.md`.

Thank you for contributing!


ShumaKitayama:
Thanks for investigating this and for adding the original regression coverage. I reproduced #4451 on the current `master` (`d75fcd4`) and tested a slightly different normalization:

```go
file := path.Clean("/" + c.Param("filepath"))
```

I think retaining the leading slash is important. `http.FileServer` calls `FileSystem.Open` with a cleaned, leading-slash URL path. With the current `[1:]` normalization, Gin's preflight check calls the same custom `http.FileSystem` with a relative path, then `http.FileServer` calls it with an absolute path. A custom filesystem that works with `http.FileServer` but validates that path shape can therefore return an unexpected 404 during Gin's preflight check.

I verified this with a wrapper around `http.FS(os.DirFS(root))` that rejects names without a leading slash or names that differ from `path.Clean(name)`:

```go
type fileServerPathFS struct {
    http.FileSystem
}

func (fileSystem fileServerPathFS) Open(name string) (http.File, error) {
    if !strings.HasPrefix(name, "/") || name != path.Clean(name) {
        return nil, fs.ErrInvalid
    }
    return fileSystem.FileSystem.Open(name)
}
```

For a request to `/tutorials/making-gin/`:

- current `master`: test fails with 404 because the preflight path has a trailing slash
- this PR's `[1:]` normalization: test fails with 404 because the preflight path is relative
- preserving the leading slash: test passes and matches the path shape used by `http.FileServer`

I also ran the resulting patch against the current `master` with `go test ./...`, `go test -race ./...`, `-tags nomsgpack`, `-tags go_json`, `-tags sonic`, and `go vet ./...`; all passed.

Would you consider preserving the leading slash and extending the regression test to verify that Gin's preflight check uses the same normalized path shape as `http.FileServer`?

Issue #4457, opened 2025-12-01
Adding a route is not concurrent-safe


maxsours:
### Description

Running the following test with `-race` trips the race detector:
```
func TestRepro(t *testing.T) {
	r := gin.New()
	go r.Routes() // could also be server startup + send request
	r.GET("/", func(ctx *gin.Context) {
		ctx.String(200, "OK")
	})
	time.Sleep(time.Second)
}
```

Output:
```
==================
WARNING: DATA RACE
Read at 0x00c0001b94c8 by goroutine 8:
  github.com/gin-gonic/gin.(*Engine).Routes()
      /Users/maxsours/go/pkg/mod/github.com/gin-gonic/gin@v1.11.0/gin.go:381 +0x3c
  max-assess-app.TestRepro.gowrap1()
      /Users/maxsours/Documents/go/max-assess-app/main_test.go:20 +0x20

Previous write at 0x00c0001b94c8 by goroutine 7:
  github.com/gin-gonic/gin.(*Engine).addRoute()
      /Users/maxsours/go/pkg/mod/github.com/gin-gonic/gin@v1.11.0/gin.go:365 +0x22c
  github.com/gin-gonic/gin.(*RouterGroup).handle()
      /Users/maxsours/go/pkg/mod/github.com/gin-gonic/gin@v1.11.0/routergroup.go:89 +0x14c
  github.com/gin-gonic/gin.(*RouterGroup).GET()
      /Users/maxsours/go/pkg/mod/github.com/gin-gonic/gin@v1.11.0/routergroup.go:117 +0x104
  max-assess-app.TestRepro()
      /Users/maxsours/Documents/go/max-assess-app/main_test.go:21 +0x94
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1934 +0x164
  testing.(*T).Run.gowrap1()
      /usr/local/go/src/testing/testing.go:1997 +0x3c
```

I believe this is due to a lack of concurrent safety when accessing `gin.Engine.trees`. One goroutine iterates through `trees` while another goroutine calls appends to it. The analogous operation on `http.ServeMux` is concurrent-safe due to their use of a `sync.RWMutex` when adding/reading new route patterns: 
https://cs.opensource.google/go/go/+/refs/tags/go1.25.4:src/net/http/server.go;l=2940

### Gin Version

v1.11.0

### Can you reproduce the bug?

Yes

### Source Code

Run the following test with `-race`:
```
func TestRepro(t *testing.T) {
[...]

PR #4459, opened 2025-12-02
Fix: make Engine.Routes() concurrency-safe (#4457)

S-Sarim:
# fix: make Engine.Routes() concurrent-safe (#4457)

## Description
This PR addresses a **data race in `Engine.Routes()`** that occurs when routes are read concurrently with route registration.

### Problem
- Concurrent calls to `Engine.Routes()` while registering new routes could cause **simultaneous reads and writes** to the internal `trees` structure.  
- This leads to **race conditions**, which are detectable with `go test -race`.  
- Example scenario:

```go
r := gin.New()
go r.Routes()          // concurrent read
r.GET("/", handler)    // concurrent write

```

### Solution
-Introduced a sync.RWMutex in Engine to protect access to the route trees.
-Reads (Routes()) acquire a read lock, while route registrations acquire a write lock.
-Ensures thread-safe access to route trees without blocking unrelated reads.
-Added a unit test (TestRoutesConcurrent) to reproduce the race condition and verify the fix.

### Testing
-Run go test ./... -race locally; the previous race warnings no longer appear.
-The test confirms that concurrent reads and writes to routes are handled safely.

## References
-Fixes issue: #4457

### DIFF
--- a/engine_test.go
+++ b/engine_test.go
@@ -1,0 +1,22 @@
+package gin
+
+import (
+	"testing"
+)
+
+func TestRoutesConcurrent(t *testing.T) {
+	r := New()
+
+	done := make(chan bool)
+
+	// Concurrently read routes
+	go func() {
+		_ = r.Routes()
+		done <- true
+	}()
+
+	// Register a route at the same time
+	r.GET("/", func(c *Context) { c.String(200, "OK") })
+
+	<-done
+}
--- a/gin.go
+++ b/gin.go
@@ -182,6 +182,7 @@ type Engine struct {
 	noMethod         HandlersChain
 	pool             sync.Pool
 	trees            methodTrees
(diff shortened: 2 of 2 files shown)


appleboy:
Why is the route registration needed for concurrency-safe?

S-Sarim:
> Why is the route registration needed for concurrency-safe?

Route registration modifies the route tree, and Routes() reads from it, this causes data race, to avoid it we need concurrency safe access, that is why route registration is part of the test to ensure reads and writes don’t conflict.

appleboy:
What I mean is: why would we ever need to write code like this? In what situation would your example be used?

```go
r := gin.New()
go r.Routes()          // concurrent read
r.GET("/", handler)    // concurrent write
```

We wouldn’t normally expect this kind of usage.

PR #4472, opened 2025-12-13
fix(context): ClientIP handling for multiple X-Forwarded-For header values

### LINKED ISSUE #4468: ClientIP from RemoteIPHeaders calculation should concatenate all header values
### Description

As of now, the[ `ClientIP` function](https://github.com/gin-gonic/gin/blob/19b877fa50cbbb9282763099fb177a1e5cc5c850/context.go#L972C19-L1007) parses each header in `c.engine.RemoteIPHeaders` by calling `c.requestHeader(headerName)`, which calls [`c.Request.Header.Get(key)`](https://github.com/gin-gonic/gin/blob/19b877fa50cbbb9282763099fb177a1e5cc5c850/context.go#L1033-L1035). The official Golang docs state that `(h Header) Get(key string) string` "gets the first value associated with the given key" (https://pkg.go.dev/net/http#Header.Get).

However, consider this scenario with multiple `X-Forwarded-For` headers being set:

```
X-Forwarded-For: 1.2.3.4, 127.0.0.1
X-Forwarded-For: 5.6.7.8
```

Let 127.0.0.1 be the only trusted proxy. As of now, Gin would take the first header value, `X-Forwarded-For: 1.2.3.4, 127.0.0.1`, identify 127.0.0.1 as trusted proxy and would therefore return 1.2.3.4 as client IP.

However, citing heroku/roadmap#238, when a request with multiple X-Forwarded-For headers is received by the router/proxy it should handle the values as one continuous list, as specified by the [HTTP spec](https://datatracker.ietf.org/doc/html/rfc2616#section-4.2):

> Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)]. It MUST be possible to combine the multiple header fields into one “field-name: field-value” pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma. The order in which header fields with the same field-name are received is therefore significant to the interpretation of the combined field value, and thus a proxy MUST NOT change the order of these field values when a message is forwarded.

Thus, the above headers should be treated equivalently to:

```
X-Forwarded-For: 1.2.3.4, 127.0.0.1, 5.6.7.8
```

In which case Gin would return 5.6.7.8 as client IP.

Hence, to properly handle multiple `X-Forwarded-For` header values (or any other header name specified in `RemoteIPHeaders`), Gin should first join all header values by a comma (e.g. using `strings.Join(c.Request.Header.Values(key), ",")` instead of only looking at the first header value.

### Gin Version

master

No

### Source Code

_No response_

### Go Version

_No response_

### Operating System

_No response_

Nurysso:
**Description:**

This PR addresses the behavior described in issue #4468 by updating how `ClientIP` processes headers listed in `RemoteIPHeaders`.

When multiple headers with the same name (e.g. `X-Forwarded-For`) are present, Gin previously only considered the first header value via `Header.Get`. Per the HTTP specification, such headers must be treated as a single, ordered, comma-separated list.

This change joins all header values using `Request.Header.Values(headerName)` before validation, ensuring correct client IP resolution in multi-proxy scenarios.

**Changes:**

* Combine all values of headers in `RemoteIPHeaders` before passing them to `validateHeader`.
* Add tests covering multiple header values to ensure correct behavior.

**Checklist:**

* [x] PR opened against `master`
* [x] All CI tests pass
* [x] Tests added to cover the updated behavior
* [ ] Documentation update not required (bug fix, not a new feature)

**References:**

* Fixes #4468

### DIFF
--- a/context_test.go
+++ b/context_test.go
@@ -1144,66 +1144,34 @@ func TestContextRenderNoContentIndentedJSON(t *testing.T) {
 }
 
 func TestContextClientIPWithMultipleHeaders(t *testing.T) {
-	// Create a new Gin engine
-	engine := New()
+	c, _ := CreateTestContext(httptest.NewRecorder())
+	c.Request, _ = http.NewRequest(http.MethodGet, "/test", nil)
 
-	// Set trusted proxies
-	err := engine.SetTrustedProxies([]string{"127.0.0.1"})
-	if err != nil {
-		t.Fatalf("Failed to set trusted proxies: %v", err)
-	}
-	engine.ForwardedByClientIP = true
-	engine.RemoteIPHeaders = []string{"X-Forwarded-For"}
+	// Multiple X-Forwarded-For headers
+	c.Request.Header.Add("X-Forwarded-For", "1.2.3.4, "+localhostIP)
+	c.Request.Header.Add("X-Forwarded-For", "5.6.7.8")
+	c.Request.RemoteAddr = localhostIP + ":1234"
 
-	// Create a test request with multiple X-Forwarded-For headers
-	req := httptest.NewRequest(http.MethodGet, "/test", nil)
-	req.Header.Add("X-Forwarded-For", "1.2.3.4, 127.0.0.1")
-	req.Header.Add("X-Forwarded-For", "5.6.7.8")
-	req.RemoteAddr = "127.0.0.1:1234"
+	c.engine.ForwardedByClientIP = true
+	c.engine.RemoteIPHeaders = []string{"X-Forwarded-For"}
+	_ = c.engine.SetTrustedProxies([]string{localhostIP})
 
-	// Create response recorder
-	w := httptest.NewRecorder()
-
-	// Create context
-	c, _ := CreateTestContext(w)
-	c.Request = req
-	c.engine = engine
-
-	// Test ClientIP
-	clientIP := c.ClientIP()
-
-	// Should return 5.6.7.8 (the last non-trusted IP)
-	expected := "5.6.7.8"
-	if clientIP != expected {
-		t.Errorf("Expected ClientIP to be %s, got %s", expected, clientIP)
-	}
+	// Should return 5.6.7.8 (last non-trusted IP)
+	assert.Equal(t, "5.6.7.8", c.ClientIP())
 }
 
 func TestContextClientIPWithSingleHeader(t *testing.T) {
-	engine := New()
+	c, _ := CreateTestContext(httptest.NewRecorder())
+	c.Request, _ = http.NewRequest(http.MethodGet, "/test", nil)
+	c.Request.Header.Set("X-Forwarded-For", "1.2.3.4, "+localhostIP)
+	c.Request.RemoteAddr = localhostIP + ":1234"
 
-	if err := engine.SetTrustedProxies([]string{"127.0.0.1"}); err != nil {
-		t.Fatalf("SetTrustedProxies failed: %v", err)
-	}
+	c.engine.ForwardedByClientIP = true
+	c.engine.RemoteIPHeaders = []string{"X-Forwarded-For"}
+	_ = c.engine.SetTrustedProxies([]string{localhostIP})
 
-	engine.ForwardedByClientIP = true
-	engine.RemoteIPHeaders = []string{"X-Forwarded-For"}
-
-	req := httptest.NewRequest(http.MethodGet, "/test", nil)
-	req.Header.Set("X-Forwarded-For", "1.2.3.4, 127.0.0.1")
-	req.RemoteAddr = "127.0.0.1:1234"
-
-	w := httptest.NewRecorder()
-	c, _ := CreateTestContext(w)
-	c.Request = req
-	c.engine = engine
-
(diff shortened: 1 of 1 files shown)


fxedel:
Looks good to me!

anfaas1618:
please rebase or merge with master. this pr is behind

context_test.go
```diff
@@ -1143,6 +1143,62 @@ func TestContextRenderNoContentIndentedJSON(t *testing.T) {
 	assert.Equal(t, "application/json; charset=utf-8", w.Header().Get("Content-Type"))
 }
 
+func TestContextClientIPWithMultipleHeaders(t *testing.T) {
+	// Create a new Gin engine
+	engine := New()
+
+	// Set trusted proxies
+	engine.SetTrustedProxies([]string{"127.0.0.1"})
```
anfaas1618:
```suggestion
	engine.SetTrustedProxies([]string{localhostIP})
```

context_test.go
```diff
@@ -1143,6 +1143,62 @@ func TestContextRenderNoContentIndentedJSON(t *testing.T) {
...
+	// Should return 5.6.7.8 (the last non-trusted IP)
+	expected := "5.6.7.8"
+	if clientIP != expected {
+		t.Errorf("Expected ClientIP to be %s, got %s", expected, clientIP)
+	}
+}
+
+func TestContextClientIPWithSingleHeader(t *testing.T) {
+	engine := New()
+	engine.SetTrustedProxies([]string{"127.0.0.1"})
```
anfaas1618:
```suggestion
	engine.SetTrustedProxies([]string{localhostIP})
```

Nurysso:
> please rebase or merge with master. this pr is behind

done, also regarding the changes you requested with use of localhostIP instead of hardcoded 127.0.0.1 fails the test I wrote should I leave it or do something like this

```go
func TestContextClientIPWithSingleHeader(t *testing.T) {
	c.Request, _ = http.NewRequest(http.MethodGet, "/test", nil)
	c.Request.Header.Set("X-Forwarded-For", fmt.Sprintf("1.2.3.4, %s", localhostIP))
	c.Request.RemoteAddr = fmt.Sprintf("%s:1234", localhostIP)
	
	c.engine.ForwardedByClientIP = true
	c.engine.RemoteIPHeaders = []string{"X-Forwarded-For"}
	_ = c.engine.SetTrustedProxies([]string{localhostIP})
	
	// Should return 1.2.3.4
	assert.Equal(t, "1.2.3.4", c.ClientIP())
}
```

I used AI for this and not sure if this is the best way.

appleboy:
@Nurysso, please take a look at the CI testing failure.

appleboy:
@anfaas1618 Please help review again.

anfaas1618:
> > please rebase or merge with master. this pr is behind
> 
> done, also regarding the changes you requested with use of localhostIP instead of hardcoded 127.0.0.1 fails the test I wrote should I leave it or do something like this
> 
> ```go
> func TestContextClientIPWithSingleHeader(t *testing.T) {
> 	c, _ := CreateTestContext(httptest.NewRecorder())
> 	c.Request, _ = http.NewRequest(http.MethodGet, "/test", nil)
> 	c.Request.Header.Set("X-Forwarded-For", fmt.Sprintf("1.2.3.4, %s", localhostIP))
> 	c.Request.RemoteAddr = fmt.Sprintf("%s:1234", localhostIP)
> 	
> 	c.engine.ForwardedByClientIP = true
> 	c.engine.RemoteIPHeaders = []string{"X-Forwarded-For"}
> 	_ = c.engine.SetTrustedProxies([]string{localhostIP})
> 	
> 	// Should return 1.2.3.4
> 	assert.Equal(t, "1.2.3.4", c.ClientIP())
> }
> ```
> 
> I used AI for this and not sure if this is the best way.

this is good , you can use this, fmt.Sprintf is discouraged but since its used only for testing its ok to use this way, it will make sure we are correctly and consistently using the consts defined which is easier when writing or updating test cases.

Nurysso:
> this is good , you can use this, fmt.Sprintf is discouraged but since its used only for testing its ok to use this way, it will make sure we are correctly and consistently using the consts defined which is easier when writing or updating test cases.

Okay understood, but in this test golangci-lint fails cause Sprintf. Should i update it to use string concatenation?

anfaas1618:
> > this is good , you can use this, fmt.Sprintf is discouraged but since its used only for testing its ok to use this way, it will make sure we are correctly and consistently using the consts defined which is easier when writing or updating test cases.
> 
> Okay understood, but in this test golangci-lint fails cause Sprintf. Should i update it to use string concatenation?
``` Golang
c.Request.Header.Set("X-Forwarded-For", "1.2.3.4, " + localhostIP)
c.Request.RemoteAddr = localhostIP + ":1234"
```
this should be fine here , why 
no need for extra allocations or parsing

anfaas1618 (approved):

appleboy (approved):

appleboy:
@Nurysso @anfaas1618 Thanks All.

Issue #4188, opened 2025-03-18
About os.Chmod in SaveUploadedFile


NezhaFan:
I submitted a PR #4181 , which has been merged.   But there's one question that remains.

`os.MkdirAll` checks whether the directory exists, and if it exists, it does nothing.    But `os.Chmod` does not check whether the folder permissions are the same as currently, it will cause `syscall.Chmod` to change permissions every time.

after merged:
<img width="911" alt="Image" src="https://github.com/user-attachments/assets/79e53fad-ca7f-4113-9ef1-d8d46b56256b" />

krombel:
It is even worse: In my case my application has no right to run chmod (but permissions are right)
=> SaveUploadedFile is not usable for me anymore as it always fails

PR #4482, opened 2025-12-27
context.SaveUploadedFile: Only run Chmod if mode needs change

krombel:
# Pull Request Checklist

- [x] Tests are added or modified as needed to cover code changes. (interception of syscalls not possible natively)

Fixes the issue mentioned in https://github.com/gin-gonic/gin/issues/4188#issuecomment-3693945581


context.go
```diff
@@ -720,7 +720,12 @@ func (c *Context) SaveUploadedFile(file *multipart.FileHeader, dst string, perm
...
 	}
-	if err = os.Chmod(dir, mode); err != nil {
+	// Only run Chmod if mode is not as configured
+	info, err := os.Stat(dir)
+	if err == nil && info.Mode().Perm() != mode {
+		err = os.Chmod(dir, mode)
+	}
+	if err != nil {
 		return err
 	}
```
anfaas1618, line 741:
```suggestion
// Only run Chmod if mode is not as configured
if info, statErr := os.Stat(dir); statErr == nil && info.Mode().Perm() != mode {
    if err := os.Chmod(dir, mode); err != nil {
        return err
    }
} else if statErr != nil {
    return statErr
}
```
since this info is used in same block of error , can we scope it better,

krombel: I did it this way to not cause codecov/patch to fail...
I don't know how to see the details on codecov now but in this run it failed an my approach now did not cause this error
https://github.com/gin-gonic/gin/runs/59003196111
My original patch was https://github.com/gin-gonic/gin/commit/9ea8851a4abfd941c7e6c0969bb5502a495c3af2 which is quite similar to your approach. Codecov failed on https://github.com/gin-gonic/gin/commit/9ea8851a4abfd941c7e6c0969bb5502a495c3af2#diff-552f47512a00afe5fc6850cc9ddc830a6daeca162750e50aab4ed549685e0253R731 so the `return err` decreased codecov => fail

Sadly I do not know how to increase codecoverage in this code area through testing so either I accept your change and a failing codecov/patch check is acceptable or this approach is...

anfaas1618: > I did it this way to not cause codecov/patch to fail... I don't know how to see the details on codecov now but in this run it failed an my approach now did not cause this error https://github.com/gin-gonic/gin/runs/59003196111 My original patch was [9ea8851](https://github.com/gin-gonic/gin/commit/9ea8851a4abfd941c7e6c0969bb5502a495c3af2) which is quite similar to your approach. Codecov failed on [9ea8851#diff-552f47512a00afe5fc6850cc9ddc830a6daeca162750e50aab4ed549685e0253R731](https://github.com/gin-gonic/gin/commit/9ea8851a4abfd941c7e6c0969bb5502a495c3af2#diff-552f47512a00afe5fc6850cc9ddc830a6daeca162750e50aab4ed549685e0253R731) so the `return err` decreased codecov => fail
> 
> Sadly I do not know how to increase codecoverage in this code area through testing so either I accept your change and a failing codecov/patch check is acceptable or this approach is...

``` golang
func TestSaveUploadedFileWithModeDifferent(t *testing.T) {
	buf := new(bytes.Buffer)
	mw := multipart.NewWriter(buf)
	w, _ := mw.CreateFormFile("file", "test")
	w.Write([]byte("test"))
	mw.Close()
	c.Request, _ = http.NewRequest(http.MethodPost, "/", buf)
	c.Request.Header.Set("Content-Type", mw.FormDataContentType())
	f, _ := c.FormFile("file")

	testFile := filepath.Join("test_dir2", "test")
	t.Cleanup(func() {
		os.Remove(testFile)
		os.Remove("test_dir2")
	})

	os.MkdirAll("test_dir2", 0o700)
	assert.NoError(t, c.SaveUploadedFile(f, testFile, 0o755))
}
```
can you try this test in context_test , also test against 750 if possible, and check coverage

krombel: so far so good. My issue was, that I was not able to trigger the "return err" path so intentionally causing the case, that an error is thrown, which would cause that code path to be triggered and therefore coverage to not decrease

anfaas1618: can you add the test cases for it and clear out the coverage part, the diff is behind ⚠️ Report is 229 commits behind head on master.

but will need to add test for changes done .

PR #4487, opened 2025-12-30
test binding, add form_test.go for form bindings

miladev95:
Add `binding/form_test.go` with unit tests for Form, FormPost, and FormMultipart.
**Tests cover:** 
names, basic POST binding, multipart binding using existing helper, and ensuring Form ignores non-multipart ParseMultipartForm error.


Nurysso:
The test looks good but Gin already has similar tests covering this functionality in `binding/multipart_form_mapping_test.go`

If you'd like to contribute, consider enhancing the existing test file instead of adding a new one.

PR #4492, opened 2026-01-05
fix(context): Allow Context.Value to properly support any key type

kasidej:
# Description

Fix an inconsistency in the Context.Value method where it accepted keys of any type in its signature but internally restricted them to strings when pass to Context.Get.

# Motivation

The Context.Value method had a signature accepting any for keys, but its implementation only worked with string keys due to a type assertion when pass to Context.Get. This was inconsistent with Context.Set which properly support any comparable key type. I using custom types as context keys (a common Go pattern to avoid key collisions) would find that Value didn't work while Set did.

# Changes

- modified Context.Value method in context.go
- add a test using a custom struct type in context_test.go

# Pull Request Checklist

### DIFF
--- a/context.go
+++ b/context.go
@@ -1436,10 +1436,8 @@ func (c *Context) Value(key any) any {
 	if key == ContextKey {
 		return c
 	}
-	if keyAsString, ok := key.(string); ok {
-		if val, exists := c.Get(keyAsString); exists {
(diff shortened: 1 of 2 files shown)


Nurysso:
Looks good to me
@appleboy Can you please review it.

PR #4498, opened 2026-01-13
fix(form): correctly differentiate between nil / present-but-empty slices

NyaaaWhatsUpDoc:
This fixes the regression introduced in v1.11.0 which stopped differentiating between nil and zero length (but *not nil*) slices when binding form values.

# Pull Request Checklist

Thank you for contributing!

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2014 Manuel Martínez-Almeida
+Copyright (c) 2014-present Manuel Martínez-Almeida
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ GO_VERSION=$(shell $(GO) version | cut -c 14- | cut -d' ' -f1 | cut -d'.' -f2)
 PACKAGES ?= $(shell $(GO) list ./...)
 VETPACKAGES ?= $(shell $(GO) list ./... | grep -v /examples/)
 GOFILES := $(shell find . -name "*.go")
-TESTFOLDER := $(shell $(GO) list ./... | grep -E 'gin$$|binding$$|render$$' | grep -v examples)
+TESTFOLDER := $(shell $(GO) list ./... | grep -E 'gin$$|ginS$$|binding$$|render$$' | grep -v examples)
 TESTTAGS ?= ""
 
 .PHONY: test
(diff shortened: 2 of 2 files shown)


binding/form_mapping.go
```diff
@@ -231,19 +231,24 @@ func setByForm(value reflect.Value, field reflect.StructField, form map[string][
...
 				return false, nil
 			}
 
-			vs = []string{opt.defaultValue}
 			// pre-process the default value for multi if present
 			cfTag := field.Tag.Get("collection_format")
 			if cfTag == "" || cfTag == "multi" {
 				vs = strings.Split(opt.defaultValue, ",")
+			} else {
+				vs = []string{opt.defaultValue}
```
NyaaaWhatsUpDoc, line 263: this isn't part of the fix (nor is it in the array part below), but it just means `vs` only needs to get written to once instead of twice in the case of `cfTag == "" | cfTag == "multi"`

appleboy:
@NyaaaWhatsUpDoc Please fix the conflicts.

NyaaaWhatsUpDoc:
@appleboy merge conflicts fixed!

binding/form_mapping_test.go
```diff
@@ -915,7 +915,7 @@ func TestMappingCustomSliceOfSliceUnmarshalTextDefault(t *testing.T) {
 	var s struct {
 		FileData []customPathUnmarshalText `form:"path,default=bar/foo;bar/foo/spam,parser=encoding.TextUnmarshaler" collection_format:"csv"`
 	}
-	err := mappingByPtr(&s, formSource{"path": {}}, "form")
+	err := mappingByPtr(&s, formSource{"path": nil}, "form")
```
NyaaaWhatsUpDoc, line 918: previously this was relying on an empty (but not nil!) value slice to rely on default, but an empty slice isn't the zero value. a nil slice is.

binding/form_mapping_test.go
```diff
@@ -994,7 +994,7 @@ func TestMappingCustomArrayOfArrayUnmarshalTextDefault(t *testing.T) {
 	var s struct {
 		FileData []objectIDUnmarshalText `form:"ids,default=664a062ac74a8ad104e0e80e;664a062ac74a8ad104e0e80f,parser=encoding.TextUnmarshaler" collection_format:"csv"`
 	}
-	err := mappingByPtr(&s, formSource{"ids": {}}, "form")
+	err := mappingByPtr(&s, formSource{"ids": nil}, "form")
```
NyaaaWhatsUpDoc, line 997: previously this was relying on an empty (but not nil!) value slice to rely on default, but an empty slice isn't the zero value. a nil slice is.

themavik (commented):
Switching the slice branch from len(vs)==0 to vs==nil changes how formSource with an empty slice behaves versus a missing key for defaults in setByForm; TestMappingEmptyValues now encodes that. nit: a one-line note on that reflect.Slice branch saves the next person from bisecting intent.

PR #4524, opened 2026-02-05
Fix concurrent-safe route registration (data race in addRoute/Routes)

mehrdadbn9:
This fixes a data race between \`addRoute()\` and \`Routes()\` functions where one goroutine appends to \`trees\` slice while another reads from it.

**Root Cause:**
- \`addRoute()\` reads \`engine.trees\` to check if method exists, then appends to the slice
- \`Routes()\` iterates over \`engine.trees\` to collect all routes
- When these run concurrently, it causes a data race

**Solution:**
- Added \`mu sync.RWMutex\` field to \`Engine\` struct
- \`addRoute()\` uses \`Lock()\` to ensure exclusive write access
- \`Routes()\` uses \`RLock()\` to allow concurrent reads

**Testing:**
- Original repro test from issue passes without race detector warnings
- Full test suite passes with \`-race\` flag

**Related Code:**
- \`http.ServeMux\` uses similar pattern with \`sync.RWMutex\` for concurrent safety

**Fixes:** #4457

### DIFF
--- a/logger_test.go
+++ b/logger_test.go
@@ -329,6 +329,7 @@ func TestColorForLatency(t *testing.T) {
 	assert.Equal(t, white, colorForLantency(time.Millisecond*20), "20ms should be white")
 	assert.Equal(t, green, colorForLantency(time.Millisecond*150), "150ms should be green")
 	assert.Equal(t, cyan, colorForLantency(time.Millisecond*250), "250ms should be cyan")
+	assert.Equal(t, blue, colorForLantency(time.Millisecond*400), "400ms should be blue")
 	assert.Equal(t, yellow, colorForLantency(time.Millisecond*600), "600ms should be yellow")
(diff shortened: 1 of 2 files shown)


mehrdadbn9:
@codecov-bot compare against d7776de not 3dc1cd6. The PR is based on current master which is at d7776de (feat(render): add bson protocol).

mehrdadbn9:
"Hi maintainers! 👋

All CI checks are now passing:
- ✅ Lint
- ✅ All tests (Ubuntu/macOS, Go 1.24/1.25, including race detector)
- ✅ Codecov/project
- ✅ CodeQL
- ✅ Trivy Security Scan\n\nThis PR fixes a data race between \`addRoute()\` and \`Routes()\` by adding proper mutex synchronization. The fix has been tested with the race detector enabled.\n\nWould appreciate a review when you have time. Thank you for your work on gin! 🙏"

PR #3933, opened 2024-04-25
feat(binding): Support custom BindUnmarshaler for binding.

dkkb:
## Feature
Add `binding.BindUnmarshaler` interface to support custom unmarshaling (from the suggestion in https://github.com/gin-gonic/gin/issues/2673)

## Description
This PR is inspired by the information provided on the following page: [labstack/echo/v4#BindUnmarshaler ](https://pkg.go.dev/github.com/labstack/echo/v4#BindUnmarshaler) and  [PR-3045](https://github.com/gin-gonic/gin/pull/3045).

As [ssoroka](https://github.com/ssoroka) (the author of [PR-3045](https://github.com/gin-gonic/gin/pull/3045)) mentioned:
> This is way off my radar (it's been 2 years and I've switched to huma). Feel free to take over the change and apply it in a new PR."

Therefore, I would like to open a new pull request to implement this feature. However, I have identified two potential issues with PR-3045:

1. PR-3045 also introduces support for the `encoding.TextUnmarshaler` interface. I believe there might be compatibility issues with this approach. Instead, I suggest that users implement the new interface binding.BindUnmarshaler to achieve the desired custom unmarshaler functionality.

2. In order to support arrays and slices, PR-3045 uses `vs[0]` to retrieve the first item. This approach may be susceptible to boundary issues. Upon closer examination of the implementation, I found that addressing this complexity might be challenging. As a temporary solution, I have decided not to support this functionality for arrays and slices. If there is a future demand from other users, we can consider implementing it at that time.

```go
if ok, err := trySetCustom(vs[0], value); ok || err != nil {
    return ok, err
}
```
## Usage
```go
package main

import (
  "github.com/gin-gonic/gin"
  "strings"
)

type Birthday string

func (b *Birthday) UnmarshalParam(param string) error {
  *b = Birthday(strings.Replace(param, "-", "/", -1))
  return nil
}

func main() {
  route := gin.Default()
  var request struct {
    Birthday Birthday `form:"birthday"`
  }
  route.GET("/test", func(ctx *gin.Context) {
    _ = ctx.BindQuery(&request)
    ctx.JSON(200, request.Birthday)
  })
  route.Run(":8088")
}
```

Test it with:

```sh
curl 'localhost:8088/test?birthday=2000-01-01'
```
Result
```sh
"2000/01/01"
```

## Testing
1. Unit testing, see `binding/form_mapping_test.go`
2. Integration testing, see `gin_test.go`

## References
[...]

PR #4540, opened 2026-02-17
feat(binding): auto-bind types implementing encoding.TextUnmarshaler

### LINKED ISSUE #2423: Binding uuid.UUID using ShouldBind
## Description

Binding uuid.UUID using ShouldBind gives error ["45e1f85e-bca5-458d-bd9c-c56edd8f847b"] is not valid value for uuid.UUID

## How to reproduce
Used Postman to make request:
Content-Type: multipart/form-data
Body with form-data:
key: colorID
value: 45e1f85e-bca5-458d-bd9c-c56edd8f847b

```
// Go model in which I want to map
type Test struct {
     colorID uuid.UUID `binding:"required" form:"colorID"`
}

func test(ctx *gin.Context) {
    t := Test{}
    err := ctx.ShouldBind(&t)
    if err != nil {
	panic(err)
    }
}
```

## Actual result

```
["45e1f85e-bca5-458d-bd9c-c56edd8f847b"] is not valid value for uuid.UUID
```

veeceey:
## Problem

Using `ShouldBind`, `ShouldBindQuery`, or `BindUri` with struct fields of type `uuid.UUID` (or any type backed by a fixed-size byte array implementing `encoding.TextUnmarshaler`) fails with an unhelpful error:

```
["45e1f85e-bca5-458d-bd9c-c56edd8f847b"] is not valid value for uuid.UUID
```

This has been an open issue for years and comes up constantly when building REST APIs with gin — pretty much every API deals with UUIDs at some point.

The current workaround is either:
- Use `string` and manually `uuid.Parse()` in every handler (defeats the purpose of binding)
- Add `parser=encoding.TextUnmarshaler` to every struct tag (non-obvious, undocumented for most users)

## Fix

`trySetCustom` now checks for `encoding.TextUnmarshaler` as a fallback after `BindUnmarshaler`. This means any type that implements `UnmarshalText` — `uuid.UUID`, `net.IP`, custom types, etc. — works automatically with form, query, and URI binding. No extra struct tags needed.

Precedence is preserved:
1. `BindUnmarshaler` (`UnmarshalParam`) — highest priority
2. `encoding.TextUnmarshaler` (`UnmarshalText`) — new automatic fallback
3. Default type-based binding (int, string, etc.)

`time.Time` is explicitly excluded from the automatic `TextUnmarshaler` path since gin already has dedicated time handling via `time_format`, `time_utc`, and `time_location` tags.

## Before / After

```go
type Request struct {
    ColorID uuid.UUID `form:"colorID" binding:"required"`
}

func handler(c *gin.Context) {
    var req Request
    // Before: fails with "is not valid value for uuid.UUID"
    // After: works correctly
    if err := c.ShouldBind(&req); err != nil {
        c.JSON(400, gin.H{"error": err.Error()})
        return
    }
    c.JSON(200, req)
}
```

## Testing

- Added 7 new tests covering UUID-like `[16]byte` types with `TextUnmarshaler`:
  - Form binding, URI binding, slice binding, multi-value, defaults, invalid input
  - Precedence test confirming `BindUnmarshaler` still wins over `TextUnmarshaler`
- Updated 1 existing test to match the new automatic binding behavior
- All existing tests pass (`go test ./...`)

Fixes #2423

### DIFF
--- a/binding/form_mapping_test.go
+++ b/binding/form_mapping_test.go
@@ -1255,8 +1255,10 @@ func (d *testDualUnmarshaler) UnmarshalText(text []byte) error {
 	return nil
 }
 
-var _ BindUnmarshaler = (*testDualUnmarshaler)(nil)
-var _ encoding.TextUnmarshaler = (*testDualUnmarshaler)(nil)
+var (
+	_ BindUnmarshaler          = (*testDualUnmarshaler)(nil)
+	_ encoding.TextUnmarshaler = (*testDualUnmarshaler)(nil)
+)
 
 func TestMappingBindUnmarshalerTakesPrecedenceOverTextUnmarshaler(t *testing.T) {
 	var s struct {


takanuva15:
@veeceey I [wrote the PR](https://github.com/gin-gonic/gin/pull/4203) for adding `encoding.TextUnmarshaler` support. 

- The new parser tag is fully documented in `doc.md`. When gin publishes 1.12 and updates their doc page, everyone will see it.
- Your change is not backwards-compatible. There are thousands of existing Gin servers out there that could start throwing errors because they will suddenly start using `encoding.UnmarshalText` method instead of Gin's default logic.  This compatibility issue with `encoding.TextUnmarshaler` was mentioned [in an earlier PR](https://github.com/gin-gonic/gin/pull/3933) too. 

While I would like using `encoding.TextUnmarshaler` by default, I don't know if the maintainers are willing to break backwards-compatibility for this. Just fyi

veeceey:
Thanks for the heads up @takanuva15, and nice work on #4203! You raise a valid concern about backwards compatibility. I hadn't realized the TextUnmarshaler support was already being added through the parser tag approach.

Since your PR covers this in a backwards-compatible way with the explicit tag, I think it makes sense to defer to that approach. I'm happy to close this PR if the maintainers prefer the opt-in parser tag route. Let me know what you all think.

veeceey:
Closing this - as pointed out, #4203 already covers TextUnmarshaler support via the parser tag approach, which is more backwards-compatible. Thanks for the review!

PR #4569, opened 2026-03-07
feat(engine): add H2CConfig to allow configuring h2c http2.Server

### LINKED ISSUE #3410: Engine.Handler()  should initialize HTTP2 server Read/WriteTimeout to adequate Values
## Description

`Engine.Handler()` initialises `http2.Server` without setting `ReadTimeout` and `WriteTimeout` to adequate values, which leaves them to their default **off**  values. 

This expose the http server to [slowloris](https://www.cloudflare.com/en-gb/learning/ddos/ddos-attack-tools/slowloris/) DDoS attacks when dealing with untrusted clients.

You would be leaking connections and run out of descriptors..

## How to reproduce

You start a new engine with any of the `Run` methods. 

```
(...)
gin.SetMode(gin.ReleaseMode)
router = gin.New()
router.RunTLS("some ost url", "some ssl certificate",  "some ssl key")
(...)
```

## Expectations

The method `Engine.Handler()` should initialise the http server with proper timeout values so it doesn't expose it to the issue explained above.

i.e.

```
server := &http2.Server{
  (...),
  ReadTimeout:  5 * time.Second,
  WriteTimeout: 10 * time.Second,
  (...),
}
```

## Actual result
```
func (engine *Engine) Handler() http.Handler {
	if !engine.UseH2C {
		return engine
	}

	h2s := &http2.Server{}
	return h2c.NewHandler(engine, h2s)
}
```

## Environment

- go version: `go1.19.1`
- gin version (or commit ref): `github.com/gin-gonic/gin v1.8.1`
- operating system: `Linux bach 4.9.0-16-amd64 #1 SMP Debian 4.9.272-2 (2021-07-19) x86_64 GNU/Linux`

abhiyankhanal:
## Summary
- Adds `H2CConfig *http2.Server` field to `Engine` struct, allowing users to configure the HTTP/2 server used for H2C connections
- When `H2CConfig` is nil, the existing behavior is preserved (bare `&http2.Server{}`)
- When set, users can configure `IdleTimeout`, `MaxConcurrentStreams`, etc. to protect against slowloris attacks

## Related Issue
Closes #3410

## Changes Made
- `gin.go`: Added `H2CConfig` field to `Engine` struct with documentation
- `gin.go`: Modified `Handler()` to use `H2CConfig` when non-nil
- `gin_test.go`: Added `TestH2CHandlerDefaultConfig` and `TestH2CHandlerWithCustomConfig` tests

## Testing
- `go test -run TestH2C -v -count=1` — all tests pass
- Default behavior (nil config) verified unchanged
- Custom config with `IdleTimeout` verified via end-to-end H2C request

## Example Usage
```go
engine := gin.New()
engine.UseH2C = true
engine.H2CConfig = &http2.Server{
    IdleTimeout: 30 * time.Second,
}
```


Qodo-Free-For-OSS:
Hi, TestH2CHandlerWithCustomConfig reads from res.Body but never closes it, leaking connections/file descriptors and potentially making the test suite flaky under parallel runs.

**Severity:** action required | **Category:** reliability

**How to fix:** Close response body in test

> ### Issue description
> `TestH2CHandlerWithCustomConfig` does not close the HTTP response body, leaking resources and risking test flakiness.
> 
> ### Issue Context
> `http.Client.Get` returns a response whose `Body` must be closed.
> 
> ### Fix Focus Areas
> - gin_test.go[163-170]
>   - After a successful `Get`, add `defer res.Body.Close()` (or close after reading).

---

appleboy:
Thanks for the PR — the direction is solid and the API surface is clean and backward-compatible. I have a few concerns before this is ready to merge, mostly around the tests not actually exercising the new code path.

**1. `TestH2CHandlerDefaultConfig` doesn't test the new behavior**

```go
func TestH2CHandlerDefaultConfig(t *testing.T) {
    r := Default()
    r.UseH2C = true
    handler := r.Handler()
    assert.NotEqual(t, r, handler)
}
```

This test passes both before and after the change — it only asserts that `Handler()` returns a wrapper when `UseH2C = true`, which is pre-existing behavior already covered indirectly by `TestH2c`. It doesn't touch `H2CConfig` at all. I'd suggest removing it, or replacing it with something that actually verifies the nil path is taken.

**2. `TestH2CHandlerWithCustomConfig` doesn't verify `H2CConfig` is applied**

The test sets `IdleTimeout: 60 * time.Second` and then issues a normal GET request. The request would succeed identically with the default `&http2.Server{}`, so this test would still pass even if the `if engine.H2CConfig != nil { h2s = engine.H2CConfig }` block were removed. It proves the config doesn't break things, but not that it's actually being used.

A lightweight fix would be to extract the server selection into a small helper:

```go
func (engine *Engine) h2cServer() *http2.Server {
    if engine.H2CConfig != nil {
        return engine.H2CConfig
    }
    return &http2.Server{}
}
```

Then the test can assert pointer identity (`engine.h2cServer() == engine.H2CConfig`), which directly verifies the config is wired through. Alternatively, a behavioral test using a very short `IdleTimeout` would work, but that risks flakiness.

**3. Minor: goroutine using `t` after test return**

```go
go func() {
    err := http.Serve(ln, r.Handler())
    if err != nil {
        t.Log(err)
    }
}()
```

Calling `t.Log` from a goroutine that may outlive the test function is a race. The existing `TestH2c` has the same pattern, so feel free to defer this, but it'd be nice to use an error channel or just drop the log since `ln.Close()` will always produce an expected error here.

**4. Minor: godoc could be more useful**

The current doc comment doesn't mention that `H2CConfig` is silently ignored unless `UseH2C = true` is also set, and it doesn't capture the actual motivation (configuring timeouts/limits to mitigate slowloris-style attacks, per #3410). Worth adding both — the PR description has the right framing, the godoc just needs to inherit it.

---

Summary: the production code change is good. The main thing I'd like to see addressed is that neither test fails if the new code path is removed, so we don't have real coverage of the feature. Happy to approve once the test gap is closed.

Issue #4572, opened 2026-03-09
Non-standard X-Forwarded-For header content is not supported


dusk5213:
### Description

When a user accesses a Gin-based web application through a reverse proxy server, and the application uses c.GetClientIP() to obtain the client IP address, it currently only recognizes the following formats:

IPv4 addresses

IPv6 addresses without square brackets

For example, the following addresses can be correctly recognized:

```
192.168.8.39
240e:318:2f4a:de56::240
```

However, in real production environments, some reverse proxy servers do not provide IP addresses in this “plain” format. For example, IIS automatically adds square brackets around IPv6 addresses, which may result in the following X-Forwarded-For header:

`X-Forwarded-For: [240e:318:2f4a:de56::240]`

This format with square brackets cannot currently be recognized.

In addition, some scenarios include the client port in the X-Forwarded-For header. For example:

When IIS uses Application Request Routing (ARR) + URL Rewrite as a reverse proxy and the “Include TCP port from clientIP” option is enabled.

Some cloud load balancers also provide similar settings, such as
routing.http.xff_client_port.enabled.

In these cases, the X-Forwarded-For header may look like this:

```
X-Forwarded-For: 192.168.8.39:38792
X-Forwarded-For: [240e:318:2f4a:de56::240]:38792
```

These formats that include a port number are also not recognized.

### Gin Version

v1.12.0

### Can you reproduce the bug?

Yes

### Source Code

``` go
// main.go
package main
import (
	"github.com/gin-gonic/gin"
)
func main() {
	r := gin.Default()
	r.SetTrustedProxies([]string{"0.0.0.0/0", "::/0"})
	r.GET("/ip", func(c *gin.Context) {
		c.String(200, c.ClientIP())
	})
	r.Run(":8081")
}
```

Execute the test in Windows CMD, Test results：
``` cmd

curl -H "X-Forwarded-For: 192.168.8.39" http://127.0.0.1:8080/ip
Response content: 192.168.8.39

curl -H "X-Forwarded-For: 240e:318:2f4a:de56::240" http://127.0.0.1:8080/ip
Response content: 240e:318:2f4a:de56::240

curl -H "X-Forwarded-For: [240e:318:2f4a:de56::240]" http://127.0.0.1:8080/ip
Response content: 127.0.0.1

curl -H "X-Forwarded-For: 192.168.8.39:38792" http://127.0.0.1:8080/ip
Response content: 127.0.0.1

[...]

PR #4575, opened 2026-03-10
fix: Non-standard X-Forwarded-For header content is not supported

dusk5213:
This PR fixes the issue reported in https://github.com/gin-gonic/gin/issues/4572.

Changes include:

**1. Support IPv6 addresses with brackets**
- Previously, c.GetClientIP() could not recognize addresses like [240e:318:2f4a:de56::240]. They are now parsed correctly.

**2. Support IPv4/IPv6 addresses with port numbers**
- Fixed the issue where X-Forwarded-For headers containing ports, common in IIS, ARR, or cloud load balancer scenarios, could not be recognized, e.g.:
```
127.0.0.1:38792
[240e:318:2f4a:de56::240]:38792
```
**3. Preserve original behavior**
- Plain IPv4 and IPv6 addresses are still recognized as before.

**4. Add regression tests**
- Covers the above problem cases as well as normal cases to ensure the fix works and does not break existing functionality.
This change only affects the parsing logic in validateHeader and does not impact other features.

### DIFF
--- a/gin.go
+++ b/gin.go
@@ -483,9 +483,33 @@ func (engine *Engine) validateHeader(header string) (clientIP string, valid bool
 	if header == "" {
 		return "", false
 	}
+	var ipStr string
 	items := strings.Split(header, ",")
 	for i := len(items) - 1; i >= 0; i-- {
-		ipStr := strings.TrimSpace(items[i])
+		item := strings.TrimSpace(items[i])
+		// [IPv6] or [IPv6]:port
+		if strings.HasPrefix(item, "[") {
+			if idx := strings.IndexByte(item, ']'); idx > 0 {
+				ipPart := item[1:idx]
+				if ip := net.ParseIP(ipPart); ip != nil {
+					if idx == len(item)-1 || (idx+1 < len(item) && item[idx+1] == ':') {
+						ipStr = ipPart
+					}
+				}
+			}
+		} else if ip := net.ParseIP(item); ip != nil {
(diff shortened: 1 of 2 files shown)


Qodo-Free-For-OSS:
Hi, `validateHeader` does not reset `ipStr` per comma-separated element, so an invalid element can be ignored if a later (rightmost) element already set `ipStr`, causing syntactically-invalid `X-Forwarded-For` headers to be treated as valid.

**How to fix:** Reset ipStr per item

> ### Issue description
> `validateHeader` can ignore malformed elements in a comma-separated `X-Forwarded-For` list because `ipStr` is not reset at the start of each loop iteration.
> 
> ### Issue Context
> - `ClientIP()` is documented to fall back when headers are not syntactically valid.
> - Current loop can continue with a stale `ipStr` when an element fails parsing.
> 
> ### Fix Focus Areas
> - gin.go[486-523]
>   - Set `ipStr = ""` at the top of each `for` iteration.
>   - Ensure the function rejects the header when the *current* item is invalid (do not validate using a stale prior value).
> - gin_test.go[1088-1188]
>   - Add regression tests such as:
>     - `"1.1.1.1, abc, 2.2.2.2"` -> invalid
>     - `"1.1.1.1, [240e:...]:123, bad"` -> invalid

---
*Spotted by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review - free for open-source projects.*

PR #4627, opened 2026-04-09
fix: race condition in Copy(), invalid AsciiJSON Unicode, Proxy-Authorization leak

barry3406:
## Summary

Three independent bug fixes found during code audit.

### 1. Race condition in `Context.Copy()` (`context.go`)

`Copy()` reads `c.Keys` **before** acquiring the read lock, then clones the stale reference under the lock:

```go
cKeys := c.Keys      // reads WITHOUT lock
c.mu.RLock()
cp.Keys = maps.Clone(cKeys)  // clones potentially stale reference
c.mu.RUnlock()
```

If `Set()` is called concurrently between the read and the lock, `cKeys` may reference a map being modified. This is especially dangerous on the first `Set()` call where `c.Keys` transitions from `nil` to a new map.

**Fix**: Move the `c.Keys` read inside the `RLock`.

### 2. `AsciiJSON` produces invalid JSON for supplementary Unicode (`render/json.go`)

Characters above U+FFFF (emoji, math symbols) are escaped as `\u1f600` (5+ hex digits), which is **invalid JSON** per RFC 8259. Strict JSON parsers reject this output.

Per the spec, supplementary plane characters must be encoded as UTF-16 surrogate pairs (e.g., U+1F600 → `\uD83D\uDE00`).

**Fix**: Detect runes > 0xFFFF and encode as surrogate pairs.

### 3. `Proxy-Authorization` header leaked in recovery panic logs (`recovery.go`)

`secureRequestDump` only masks `Authorization` but not `Proxy-Authorization`, which also carries credentials (used by gin's own `BasicAuthForProxy` middleware). When a panic occurs behind proxy auth, credentials are logged in plaintext.

**Fix**: Also sanitize `Proxy-Authorization` in the same check.

### DIFF
--- a/context.go
+++ b/context.go
@@ -132,9 +132,8 @@ func (c *Context) Copy() *Context {
 	cp.handlers = nil
 	cp.fullPath = c.fullPath
 
-	cKeys := c.Keys
 	c.mu.RLock()
-	cp.Keys = maps.Clone(cKeys)
+	cp.Keys = maps.Clone(c.Keys)
 	c.mu.RUnlock()
 
 	cParams := c.Params
--- a/context.go
+++ b/context.go
@@ -133,7 +133,9 @@ func (c *Context) Copy() *Context {
 	cp.fullPath = c.fullPath
 
 	c.mu.RLock()
-	cp.Keys = maps.Clone(c.Keys)
+	if c.Keys != nil {
+		cp.Keys = maps.Clone(c.Keys)
+	}
 	c.mu.RUnlock()
 
 	cParams := c.Params
--- a/recovery.go
+++ b/recovery.go
@@ -99,8 +99,8 @@ func secureRequestDump(r *http.Request) string {
 	httpRequest, _ := httputil.DumpRequest(r, false)
 	lines := strings.Split(bytesconv.BytesToString(httpRequest), "\r\n")
 	for i, line := range lines {
-		if strings.HasPrefix(line, "Authorization:") {
-			lines[i] = "Authorization: *"
+		if strings.HasPrefix(line, "Authorization:") || strings.HasPrefix(line, "Proxy-Authorization:") {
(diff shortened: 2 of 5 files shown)


context.go
Nurysso, line 1: I think a adding a nil check would be better
```go
// - cKeys := c.Keys
    c.mu.RLock()
    if c.Keys != nil {
        cp.Keys = maps.Clone(c.Keys)
    }
    c.mu.RUnlock()
```

recovery.go
Nurysso, line 1: Few things I noticed 
- Case sensitivity issue: HTTP headers are case-insensitive, but the code only matches exact case. A header like authorization: or AUTHORIZATION: would leak. Use strings.EqualFold or convert to lowercase before comparison.
- Compared to the original implementation, this adds extra complexity (SplitN) without a clear benefit. Since we already know the header name from the prefix check, a simple hardcoded replacement (as before) seems easier to read and maintain.

Nurysso (changes requested):
LGTM for the most parts, but a few minor suggestions

Nurysso (commented):
LGTM

barry3406:
Both addressed — thanks for the careful review.

- context.go: the earlier commit already moved to the `if c.Keys != nil { cp.Keys = maps.Clone(c.Keys) }` form under the same RLock, matching your suggestion exactly.
- recovery.go: dropped the small `hasHeaderPrefixFold` helper and inlined the `len + strings.EqualFold` check in `secureRequestDump`, so it's back to a simple hardcoded replacement per header, just case-insensitive. Also tightened the existing lowercase/mixed-case Authorization tests to assign directly into `r.Header` (bypassing `http.Header.Set` canonicalization) so they actually exercise the wire-level case-insensitive path, and added uppercase Authorization plus lowercase/uppercase Proxy-Authorization cases.

PR #4629, opened 2026-04-11
fix(context): skip chmod on existing dir in SaveUploadedFile

### LINKED ISSUE #4622: SaveUploadedFile: unexpected chmod on existing directories (breaks /tmp usage)
## Description

After upgrading from v1.10.1 to v1.12.0, `Context.SaveUploadedFile` started to fail when saving files directly into existing system directories like `/tmp`.

The failure happens due to an attempt to call `os.Chmod` on the target directory, even if it already exists and is not owned by the process.

Relevant code:
https://github.com/gin-gonic/gin/blob/v1.12.0/context.go#L734

## Reproduction

```go
file, _ := c.FormFile("file")

// Saving directly into /tmp
err := c.SaveUploadedFile(file, "/tmp/test.txt")
if err != nil {
  panic(err)
}
```

## Actual behavior

The call fails with:

chmod /tmp: operation not permitted

## Expected behavior

If the directory already exists, especially for system directories like `/tmp`, `SaveUploadedFile` should not attempt to modify its permissions.

The previous behavior (v1.10.1) only used `os.MkdirAll`, which does not change permissions on existing directories.

## Impact

This is a breaking change for common usage patterns such as:

* saving files to `/tmp/<filename>`
* writing into pre-existing directories managed by the OS or container runtime

## Context

This behavior appears to be introduced together with the change that allows specifying permissions for created directories (see #4068). However, applying `chmod` unconditionally also affects already existing directories.

## Suggested fix

Only apply `os.Chmod` when the directory is newly created, or make it optional.

For example:

* skip `chmod` if directory already exists
* or provide a flag to control this behavior

## Additional notes

This change is not mentioned as breaking in release notes, but it alters previously working behavior.

### Gin Version

v1.12.0

Yes

### Source Code

```go
file, _ := c.FormFile("file")

// Saving directly into /tmp
err := c.SaveUploadedFile(file, "/tmp/test.txt")
if err != nil {
   panic(err)
}
```

### Go Version

1.26.1

### Operating System

linux

barry3406:
# Pull Request Checklist

## Description

Since v1.12.0, `SaveUploadedFile` calls `os.Chmod` on the destination's parent directory unconditionally after `os.MkdirAll`, which breaks saving into pre-existing directories the process does not own (for example `SaveUploadedFile(file, "/tmp/foo.txt")` fails with `chmod /tmp: operation not permitted`).

The fix stats the directory before `MkdirAll` and only chmods when the stat returned `os.ErrNotExist`, so the configurable mode from #4088 still applies to directories `SaveUploadedFile` actually creates while existing directories are left alone.

Added `TestSaveUploadedFilePreservesExistingDirPermissions`, which creates a temp dir with mode `0o700`, saves into it with a different requested mode, and asserts the dir's mode survives and the file contents are correct. Added `TestSaveUploadedFileAppliesModeToCreatedDir` as a companion that saves into a nested directory which does not yet exist and asserts the new directory ends up with the requested mode.

Fixes #4622

### DIFF
--- a/context.go
+++ b/context.go
@@ -728,11 +728,16 @@ func (c *Context) SaveUploadedFile(file *multipart.FileHeader, dst string, perm 
 		mode = perm[0]
 	}
 	dir := filepath.Dir(dst)
+	_, statErr := os.Stat(dir)
 	if err = os.MkdirAll(dir, mode); err != nil {
 		return err
 	}
-	if err = os.Chmod(dir, mode); err != nil {
-		return err
+	// Only chmod the directory when MkdirAll actually created it; otherwise
+	// saving into a pre-existing dir we do not own (e.g. /tmp) fails. See #4622.
+	if errors.Is(statErr, os.ErrNotExist) {
+		if err = os.Chmod(dir, mode); err != nil {
+			return err
+		}
 	}
 
 	out, err := os.Create(dst)
--- a/context_test.go
+++ b/context_test.go
@@ -274,6 +274,68 @@ func TestSaveUploadedFileWithPermissionFailed(t *testing.T) {
 	require.Error(t, c.SaveUploadedFile(f, "test/permission_test", mode))
 }
 
+// TestSaveUploadedFilePreservesExistingDirPermissions is a regression test for
+// https://github.com/gin-gonic/gin/issues/4622. SaveUploadedFile must not chmod
+// a directory that already exists (e.g. /tmp), otherwise it fails with
+// "operation not permitted" when the process does not own that directory.
+func TestSaveUploadedFilePreservesExistingDirPermissions(t *testing.T) {
+	buf := new(bytes.Buffer)
+	mw := multipart.NewWriter(buf)
+	w, err := mw.CreateFormFile("file", "existing_dir_test")
+	require.NoError(t, err)
+	_, err = w.Write([]byte("existing_dir_test"))
+	require.NoError(t, err)
+	mw.Close()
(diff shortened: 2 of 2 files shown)


johnrald1 (approved):

Qodo-Free-For-OSS:
Hi, SaveUploadedFile decides to chmod based on a pre-MkdirAll Stat result, so another process can create the directory between Stat and MkdirAll and this code will still attempt to chmod it. This can reintroduce intermittent EPERM failures or unexpected permission changes in concurrent/shared-path scenarios.

**Severity:** remediation recommended | **Category:** reliability

**How to fix:** Chmod only on create

> ### Issue description
> `SaveUploadedFile` uses `os.Stat(dir)` before `os.MkdirAll` and later uses the stale `statErr` to decide whether to `os.Chmod(dir, mode)`. This creates a TOCTOU race where another process may create `dir` between the stat and mkdir, and gin will still attempt to chmod a directory it did not create.
> 
> ### Issue Context
> We want to preserve the PR’s intent (don’t chmod truly pre-existing dirs like `/tmp`), while also avoiding chmod based solely on a pre-check.
> 
> ### Fix Focus Areas
> - context.go[731-741]
>   - Adjust logic to determine whether this call created the directory (or at least avoid chmod in the raced case). Options include: create the final directory with `os.Mkdir` and chmod only when that succeeds, or create component-wise and track which components were created, or accept `Chmod` only when `statErr` is ErrNotExist *and* you can confirm creation happened in this call.

---

PR #4631, opened 2026-04-15
fix: skip chmod on existing directories in SaveUploadedFile

### LINKED ISSUE #4622: SaveUploadedFile: unexpected chmod on existing directories (breaks /tmp usage)

Herrtian:
## Summary

Fixes #4622

`SaveUploadedFile` called `os.Chmod` on the target directory unconditionally, even if it already existed. This broke file uploads to system directories like `/tmp`:

```
chmod /tmp: operation not permitted
```

## Change

Added a check before calling `os.Chmod`: only newly created directories get their permissions set. Pre-existing directories are left untouched.

```go
// Before: always chmod (breaks on /tmp, /var, etc.)
os.MkdirAll(dir, mode)
os.Chmod(dir, mode)  // fails if dir is /tmp

// After: only chmod new directories
dirExisted := dirExists(dir)
os.MkdirAll(dir, mode)
if !dirExisted {
    os.Chmod(dir, mode)
}
```

Updated `TestSaveUploadedFileWithPermission` to verify permissions on a newly created subdirectory rather than the current working directory. Added a Windows skip for the permission assertion since Windows doesn't support Unix file permissions.

## Test plan

- [x] `go build ./...` passes
- [x] `TestSaveUploadedFileWithPermission` passes
- [x] `TestSaveUploadedFileWithPermissionFailed` passes
- [x] `TestFileDescriptor` failure is pre-existing (fails on master too on Windows)

### DIFF
--- a/context.go
+++ b/context.go
@@ -715,6 +715,11 @@ func (c *Context) MultipartForm() (*multipart.Form, error) {
 	return c.Request.MultipartForm, err
 }
 
+func dirExists(path string) bool {
+	info, err := os.Stat(path)
+	return err == nil && info.IsDir()
+}
+
 // SaveUploadedFile uploads the form file to specific dst.
 func (c *Context) SaveUploadedFile(file *multipart.FileHeader, dst string, perm ...fs.FileMode) error {
 	src, err := file.Open()
@@ -728,11 +733,16 @@ func (c *Context) SaveUploadedFile(file *multipart.FileHeader, dst string, perm 
 		mode = perm[0]
 	}
 	dir := filepath.Dir(dst)
+	dirExisted := dirExists(dir)
 	if err = os.MkdirAll(dir, mode); err != nil {
 		return err
 	}
-	if err = os.Chmod(dir, mode); err != nil {
-		return err
+	// Only chmod newly created directories. Attempting to chmod
+	// pre-existing directories (e.g. /tmp) may fail with EPERM.
+	if !dirExisted {
+		if err = os.Chmod(dir, mode); err != nil {
+			return err
+		}
 	}
 
 	out, err := os.Create(dst)
--- a/context.go
+++ b/context.go
@@ -717,7 +717,10 @@ func (c *Context) MultipartForm() (*multipart.Form, error) {
 
 func dirExists(path string) bool {
 	info, err := os.Stat(path)
-	return err == nil && info.IsDir()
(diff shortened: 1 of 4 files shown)


Herrtian:
Added coverage for the existing directory path and the dir helper. Targeted upload-file tests pass locally.

Herrtian:
Expanded the dir helper to avoid the remaining partial coverage line. Targeted upload-file tests pass locally.

Herrtian:
Adjusted the directory setup path so the new helper branches are covered. Targeted upload-file tests pass locally with 100% coverage for the helper.

PR #4633, opened 2026-04-15
Fix: unwrap joined errors

### LINKED ISSUE #4237: Unwrap `joinErr` in `gin.Error()`
In go `errors` pkg can create a `joinErr` by `errors.Join()`, which can be `Unwrap()` into `[]error`.
`joinErr` also format it's print into `\n` separated string, which is similar with `gin.errorMsgs.String()` but can cause confuse, when they are used together:

```
Error #01 gin error
Error #02 service error
store error
Error #03 other error
```

In this case `service error` & `store error` is combined with a `joinErr`.
I think it will be better if we can see:

```
Error #01 gin error
Error #02 service error
Error #03 store error
Error #04 other error
```

Even though we can unwrap `joinErr` before we use `gin.Error()`, it will be better if `gin.Error()` do the unwrap for us as a internal logic.

odlev:
# Pull Request Checklist

## Description

`Context.Error()` now unwraps errors created by `errors.Join()` — each
inner error gets its own entry in `c.Errors` instead of being stored as a
single opaque multi-error.

**Before:**
```
Error #01: gin error
Error #02: service error
store error
Error #03: other error
```

**After:**
```
Error #01: gin error
Error #02: service error
Error #03: store error
Error #04: other error
```

The unwrap is recursive, so nested joins work too.
If a joined error contains a `*gin.Error`, its Type and Meta are preserved.
Empty `Unwrap()` slices fall through to the normal path.

Added 3 tests: basic join, nested join, join with typed `*gin.Error`.

Fixes #4237 - https://github.com/gin-gonic/gin/issues/4237

### DIFF
--- a/context.go
+++ b/context.go
@@ -244,14 +244,35 @@ func (c *Context) AbortWithError(code int, err error) *Error {
 /********* ERROR MANAGEMENT *********/
 /************************************/
 
+// joinedError is an interface for errors that wrap multiple inner errors,
+// such as those created by errors.Join.
+type joinedError interface {
+	Unwrap() []error
+}
+
 // Error attaches an error to the current context. The error is pushed to a list of errors.
 // It's a good idea to call Error for each error that occurred during the resolution of a request.
 // A middleware can be used to collect all the errors and push them to a database together,
 // print a log, or append it in the HTTP response.
 // Error will panic if err is nil.
+// If err is a joined error (created by errors.Join), it is unwrapped and each
+// individual error is added as a separate entry. The last *Error added is returned.
 func (c *Context) Error(err error) *Error {
 	if err == nil {
 		panic("err is nil")
+	}
+
+	// Unwrap joined errors so each one becomes a separate entry.
+	if joined, ok := err.(joinedError); ok {
+		errs := joined.Unwrap()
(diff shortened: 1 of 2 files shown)


Qodo-Free-For-OSS:
Hi, Context.Error only unwraps joined errors when the top-level error directly implements Unwrap() []error, so joined errors wrapped by other errors will be stored as a single opaque error entry instead of being expanded into separate c.Errors entries.

**How to fix:** Use errors.As for joinedError

> ### Issue description
> `Context.Error()` unwraps joined errors only when `err` itself implements `Unwrap() []error`. If the joined error is wrapped (e.g., `fmt.Errorf("...: %w", errors.Join(...))`), it will not be unwrapped and will be appended as a single opaque error.
> 
> ### Issue Context
> The function already uses `errors.As` to discover `*gin.Error` through wrapper chains; joined errors should be discovered the same way to keep behavior consistent for wrapped errors.
> 
> ### Fix Focus Areas
> - context.go[247-289]
>   - Replace the direct type assertion with an `errors.As` probe for `joinedError` (e.g., `var joined joinedError; if errors.As(err, &joined) { ... }`).
>   - Keep the existing empty-slice fallthrough behavior.

---

Issue #4460, opened 2025-12-03
http.TimeoutHandler panics when used with Gin


paul-freeman:
### Description

Gin's `responseWriter.Flush()` method performs a direct type assertion without checking if the underlying `ResponseWriter` implements `http.Flusher`. This causes a panic when Gin is used with middleware that wraps the `ResponseWriter` with a type that doesn't implement `Flusher` (e.g., `http.TimeoutHandler`).

### Expected behavior

The `Flush()` call should be a no-op when the underlying writer doesn't support flushing, similar to how `httputil.ReverseProxy` handles it.

### Actual behavior

Panic:
```
interface conversion: *http.timeoutWriter is not http.Flusher: missing method Flush
```

Stack trace points to `response_writer.go:123`:
```
(*responseWriter).Flush: w.ResponseWriter.(http.Flusher).Flush()
```

### Root cause

In [`response_writer.go`](https://github.com/gin-gonic/gin/blob/master/response_writer.go#L123):

```go
func (w *responseWriter) Flush() {
	w.ResponseWriter.(http.Flusher).Flush()
}
```

This performs an unchecked type assertion that panics if the underlying `ResponseWriter` doesn't implement `http.Flusher`.

### Suggested fix

```
func (w *responseWriter) Flush() {
	if f, ok := w.ResponseWriter.(http.Flusher); ok {
		f.Flush()
	}
}
```

This is the standard pattern used by `httputil.ReverseProxy` and other stdlib code.

### Gin Version

v1.10.1

### Can you reproduce the bug?

Yes

### Source Code

```go
package main

import (
	"net/http"
	"net/http/httputil"
	"net/url"
	"time"

	"github.com/gin-gonic/gin"
)

func main() {
	// Backend that streams data
	go func() {
		http.HandleFunc("/download", func(w http.ResponseWriter, r *http.Request) {
			for i := 0; i < 5; i++ {
				w.Write(make([]byte, 50*1024))
				if f, ok := w.(http.Flusher); ok {
					f.Flush()
				}
				time.Sleep(100 * time.Millisecond)
			}
		})
		http.ListenAndServe(":8081", nil)
	}()
	time.Sleep(100 * time.Millisecond)

	// Gin proxy wrapped with TimeoutHandler
	gin.SetMode(gin.ReleaseMode)
	r := gin.New()
	r.Use(gin.Recovery())

	backendURL, _ := url.Parse("http://localhost:8081")
[...]

PR #4639, opened 2026-04-22
fix: safe type assertions in CloseNotify() and Hijack()

### LINKED ISSUE #4638: Unsafe type assertions in Hijack() and CloseNotify() cause panic with http.TimeoutHandler
### Description

## Problem

Similar to #4460 (fixed for `Flush()` in #4479), `Hijack()` and `CloseNotify()` still use direct type assertions that panic when the underlying writer doesn't implement the interface:

```go
// panics with http.TimeoutHandler (timeoutWriter is not http.Hijacker)
return w.ResponseWriter.(http.Hijacker).Hijack()

// panics with http.TimeoutHandler (timeoutWriter is not http.CloseNotifier)
return w.ResponseWriter.(http.CloseNotifier).CloseNotify()
```
## Environment                                                                                                                                                                              
                                                                                   
- **Go version:** go1.25.0 linux/amd64                                                                                                                                                      - **Gin version:** v1.12.0                                                       
- **OS:** Linux 6.17.0 (Ubuntu) 

## Expected behavior

Same graceful degradation already implemented in `Flush()` — return error / nil instead of panicking.

### Gin Version

d3ffc9985281dcf4d3bef604cce4e662b1a327a6

Yes

### Source Code

```go                                                                                                                                                                                       
  package main                                                                                                                                                                                
                                                                   
  import (                                                                                                                                                                                    
      "net/http"                                                                   
      "net/http/httptest"                                                                                                                                                                     
      "time"
      "testing"                                                                                                                                                                               
                                                                                   
      "github.com/gin-gonic/gin"         
  )                                                         
                                                          
  func TestTimeoutHandlerHijackPanic(t *testing.T) {               
      r := gin.New()                  
      r.GET("/ws", func(c *gin.Context) {
          // Any handler that attempts to hijack the connection (e.g. WebSocket upgrade)                                                                                                      
          _, _, err := c.Writer.Hijack()                    
          if err != nil {                                                                                                                                                                     
              c.String(http.StatusInternalServerError, err.Error())                
          }                                                        
      })                                                    
                                                                                                                                                                                              
      // Wrapping with http.TimeoutHandler is common for adding global request timeouts.                                                                                                      
      // timeoutWriter (internal to net/http) does NOT implement http.Hijacker.                                                                                                               
      handler := http.TimeoutHandler(r, 5*time.Second, "timeout")                                                                                                                             
                                                                                                                                                                                              
      req := httptest.NewRequest(http.MethodGet, "/ws", nil)
      w := httptest.NewRecorder()                                                                                                                                                             
                                                                                                                                                                                              
      // Panics: interface conversion: *http.timeoutWriter is not http.Hijacker
      handler.ServeHTTP(w, req)                                                                                                                                                               
  }
```

### Go Version

go1.25.0 linux/amd64

### Operating System

Linux 6.17.0-20-generic

jassus213:
## Problem                                                                                                                                                                                  
         
Similar to #4460 (fixed for `Flush()` in #4479), `Hijack()` and `CloseNotify()` perform direct type assertions that panic when the underlying writer doesn't implement the interface.
                                                                                                                                                                                     
When gin is used with `http.TimeoutHandler` or any middleware that wraps `http.ResponseWriter` with a type that doesn't implement `http.Hijacker` or `http.CloseNotifier`, calling `Hijack()` or `CloseNotify()` causes a runtime panic:                                                                                                                                       
                                                                                                                                                                                              
  ```go                                                  
  // panics if underlying writer is not http.CloseNotifier                                                                                                                                    
  return w.ResponseWriter.(http.CloseNotifier).CloseNotify()                       
                                                            
  // panics if underlying writer is not http.Hijacker       
  return w.ResponseWriter.(http.Hijacker).Hijack()                                                                                                                                            
  ```                                                              
                                                                                                                                                                                              
## Fix                                                                                                                                                                                      
                                                                                                                                                                                              
Replace direct assertions with checked assertions - the same pattern already used in `Flush()`:                                                                                             
                                                                                                 
 - `CloseNotify()` returns `nil` when unsupported (`http.CloseNotifier` is deprecated since Go 1.11; callers should use `Request.Context().Done()`)
 -  `Hijack()` returns `errHijackNotSupported`, consistent with the existing `errHijackAlreadyWritten` pattern                                      
                                                                                                                                                                                              
The capability check in `Hijack()` now runs before mutating `w.size`, so `Written()` stays `false` on the error path.
                                                                                                                                                                                              
## Tests                                                                         
                                                                                                                                                                                              
Added `TestResponseWriterOptionalInterfaceFallbacks` with a `mockNonHijackerWriter` (implements only `http.ResponseWriter`) verifying that `Flush()`, `CloseNotify()`, and `Hijack()` all degrade gracefully without panicking. 

Closes #4638.

### DIFF
--- a/context.go
+++ b/context.go
@@ -1328,9 +1328,15 @@ func (c *Context) SSEvent(name string, message any) {
 func (c *Context) Stream(step func(w io.Writer) bool) bool {
 	w := c.Writer
 	clientGone := w.CloseNotify()
+	var requestGone <-chan struct{}
+	if c.Request != nil {
+		requestGone = c.Request.Context().Done()
+	}
 	for {
 		select {
 		case <-clientGone:
+			return true
+		case <-requestGone:
 			return true
 		default:
 			keepOpen := step(w)
--- a/context_test.go
+++ b/context_test.go
@@ -3078,6 +3078,25 @@ func TestContextStreamWithClientGone(t *testing.T) {
 	assert.Equal(t, "test", w.Body.String())
 }
 
+func TestContextStreamWithRequestContextDone(t *testing.T) {
+	w := httptest.NewRecorder()
+	c, _ := CreateTestContext(w)
+
+	reqCtx, cancel := context.WithCancel(context.Background())
+	cancel()
+	c.Request = httptest.NewRequest(http.MethodGet, "/", nil).WithContext(reqCtx)
+
+	disconnected := c.Stream(func(writer io.Writer) bool {
+		_, err := writer.Write([]byte("test"))
+		require.NoError(t, err)
+
+		return true
+	})
+
+	assert.True(t, disconnected)
+	assert.Empty(t, w.Body.String())
+}
+
 func TestContextResetInHandler(t *testing.T) {
 	w := CreateTestResponseRecorder()
 	c, _ := CreateTestContext(w)
--- a/response_writer.go
+++ b/response_writer.go
@@ -17,7 +17,10 @@ const (
 	defaultStatus = http.StatusOK
 )
 
-var errHijackAlreadyWritten = errors.New("gin: response body already written")
+var (
+	errHijackAlreadyWritten = errors.New("gin: response body already written")
+	errHijackNotSupported   = errors.New("gin: underlying ResponseWriter does not support hijacking")
+)
 
 // ResponseWriter ...
 type ResponseWriter interface {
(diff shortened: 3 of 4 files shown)


Wsuffy:
Hit this exact panic today when adding http.TimeoutHandler to our gin setup. Looked at the PR and the fix solves it - safe assertion + graceful fallback, consistent with how Flush() is handled.

Qodo-Free-For-OSS:
Hi, responseWriter.CloseNotify() now returns nil when the underlying ResponseWriter doesn’t implement http.CloseNotifier, but Context.Stream() unconditionally selects on that channel to detect disconnects. A nil channel disables that select case permanently, so Stream will not stop on client disconnect and can loop indefinitely until step() returns false.

**How to fix:** Fallback to request context

> ### Issue description
> `responseWriter.CloseNotify()` now returns `nil` if the underlying writer does not implement `http.CloseNotifier`. `Context.Stream()` unconditionally selects on `clientGone := w.CloseNotify()`. In Go, receiving from a `nil` channel is a permanently-blocked case, so the disconnect branch never triggers and `Stream()` can’t terminate on disconnect.
> 
> ### Issue Context
> This PR targets wrappers (e.g., `http.TimeoutHandler`) that may remove optional interfaces from the underlying `http.ResponseWriter`. Those same wrappers can now cause `Stream()` to spin until `step()` returns false, even if the client is gone.
> 
> ### Fix Focus Areas
> - context.go[1326-1342]
> 
> ### Implementation notes
> - Keep `CloseNotify()`’s new safe behavior.
> - Update `Context.Stream()` to handle the nil CloseNotify channel by adding a fallback disconnect signal (preferably `c.Request.Context().Done()` when `c.Request != nil`). Example structure:
>   - compute `clientGone := c.Writer.CloseNotify()`
>   - in the loop `select { case <-clientGone: return true; case <-c.Request.Context().Done(): return true; default: ... }`
>   - if `c.Request` can be nil in any code paths, guard accordingly (e.g., only add the Done() case when non-nil).

---
*Found by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review. FYI, Qodo is free for open-source.*

jassus213:
@Qodo-Free-For-OSS 

Valid point.

CloseNotify() now returns nil when http.CloseNotifier is unsupported, while Context.Stream() previously selected only on that channel. In wrapped-writer scenarios (e.g. http.TimeoutHandler), this can disable disconnect detection in Stream().

Fixed in this PR by adding a fallback to c.Request.Context().Done() (guarded with c.Request != nil) and keeping the safe CloseNotify() behavior unchanged. Added a test for this path as well.

jassus213:
@appleboy - gentle ping on safe assertions for Hijack/CloseNotify + Stream disconnect fallback. Would appreciate a review when you can.

Thanks.

PR #4714, opened 2026-06-22
fix: flush status immediately for no-body response codes

### LINKED ISSUE #4071: ctx.Status() does not affect httptest.ResponseRecorder while ctx.JSON() does
## Description

ctx.Status() does not affect httptest.ResponseRecorder while ctx.JSON() does

## How to reproduce

```
package main

import (
	"fmt"
	"github.com/gin-gonic/gin"
	"net/http"
	"net/http/httptest"
	"testing"
)

func TestName(t *testing.T) {
	recorder := httptest.NewRecorder()

	c, _ := gin.CreateTestContext(recorder)

	changeResponseCode(c)

	fmt.Println(recorder.Code)
}

func changeResponseCode(ctx *gin.Context) {
	ctx.Status(http.StatusNoContent)
}
```

## Expectations

```
Should print 204
```

## Actual result

```
Prints 200
```

## Environment

- go version: 1.23.2
- gin version (or commit ref): 1.10.0
- operating system: Ubuntu 22.04

xbrxr03:
## Problem

`ctx.Status(code)` doesn't flush the status code to the underlying `http.ResponseWriter` when no body is written. This causes `httptest.ResponseRecorder` and similar wrappers to see the default 200 instead of the actual status code.

```go
recorder := httptest.NewRecorder()
c, _ := gin.CreateTestContext(recorder)
c.Status(http.StatusNoContent)
fmt.Println(recorder.Code) // 200 — should be 204
```

Reported in #4071.

## Root Cause

`ctx.Status()` calls `c.Writer.WriteHeader(code)`, which only stores the status in gin's `responseWriter` buffer. The status gets flushed to the underlying writer when `Write()` is called (which triggers `WriteHeaderNow()`). But for status codes that never have a body (1xx, 204, 304), no `Write()` ever happens, so the status never reaches the underlying `ResponseRecorder`.

This is inconsistent with `ctx.Render()`, which already calls `WriteHeaderNow()` for no-body status codes.

## Fix

When `Status()` is called with a no-body status code (where `bodyAllowedForStatus()` returns false), immediately flush the status to the underlying writer via `WriteHeaderNow()`. This mirrors the existing behavior in `Render()` and matches user expectations.

For status codes that allow a body (e.g. 200), `Status()` continues to buffer the status until `Write()` is called, preserving the existing behavior for handlers that set a status before writing content.

## Tests

- `TestContextStatusFlushesNoBodyCodes`: verifies 100, 101, 102, 204, and 304 flush immediately
- `TestContextStatusWithBodyDoesNotFlushEarly`: verifies 200 still buffers until Write() is called

All existing tests continue to pass.

Fixes #4071

### DIFF
--- a/context.go
+++ b/context.go
@@ -1107,8 +1107,15 @@ func bodyAllowedForStatus(status int) bool {
 }
 
 // Status sets the HTTP response code.
+// For status codes that do not allow a response body (1xx, 204, 304),
+// the status is flushed immediately so that httptest.ResponseRecorder
+// and similar wrappers record the correct status without requiring a
+// subsequent Write call.
 func (c *Context) Status(code int) {
 	c.Writer.WriteHeader(code)
+	if !bodyAllowedForStatus(code) {
+		c.Writer.WriteHeaderNow()
+	}
 }
 
 // Header is an intelligent shortcut for c.Writer.Header().Set(key, value).
--- a/context_test.go
+++ b/context_test.go
@@ -1898,6 +1898,59 @@ func TestContextAbortWithStatus(t *testing.T) {
 	assert.True(t, c.IsAborted())
 }
 
+func TestContextStatusFlushesNoBodyCodes(t *testing.T) {
+	t.Run("204 No Content is flushed immediately", func(t *testing.T) {
+		w := httptest.NewRecorder()
+		c, _ := CreateTestContext(w)
+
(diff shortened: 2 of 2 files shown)


xbrxr03:
Closing this PR due to lack of maintainer engagement. I'm happy to reopen if there's interest. Thanks for the great project.

PR #4741, opened 2026-07-20
feat(binding): NewStructValidator for go-playground/validator options

### LINKED ISSUE #4733: Allow configuring the binding validator with go-playground/validator Options (e.g. WithRequiredStructEnabled)
### Feature Description

There is currently no first-class way to construct gin's default binding validator with [go-playground/validator `Option`s](https://pkg.go.dev/github.com/go-playground/validator/v10#Option) — most importantly `WithRequiredStructEnabled()`.

**Why this matters**

`WithRequiredStructEnabled()` is what makes the `required` tag apply to non-pointer struct fields. Without it, `required` is silently a no-op on any value-typed struct field — wrapper types such as a `UUID` or `Decimal` that are defined as `type X struct{...}`. So a request that omits such a field binds its zero value and passes validation, instead of failing with a 400:

```go
type CreateReq struct {
    ID  MyUUID `json:"id" binding:"required"` // MyUUID is a struct
}
// {} binds successfully — `required` never fires.
```

**Why it can't be opted into today**

The option is constructor-only. In go-playground/validator it sets a private field with no setter, so it can only be enabled via `validator.New(validator.WithRequiredStructEnabled())`:

```go
// validator_instance.go
requiredStructEnabled  bool
// options.go
func WithRequiredStructEnabled() Option { return func(v *Validate) { v.requiredStructEnabled = true } }
```

But gin's `defaultValidator.lazyinit()` hardcodes `validator.New()` with no options:

```go
// binding/default_validator.go
func (v *defaultValidator) lazyinit() {
    v.once.Do(func() {
        v.validate = validator.New()
        v.validate.SetTagName("binding")
    })
}
```

`binding.Validator.Engine()` returns the already-constructed `*validator.Validate`, which is too late — the option must be passed at construction, and there is no setter to flip it afterwards. Because `defaultValidator` (and its `validate` field) are unexported, the only way to enable the option is to replace `binding.Validator` entirely with a custom `StructValidator`, re-implementing `ValidateStruct`'s pointer/slice/array dispatch just to pass one option:

```go
type myValidator struct{ v *validator.Validate }
func (m *myValidator) ValidateStruct(obj any) error { /* copy of gin's ptr/slice/struct dispatch */ }
func (m *myValidator) Engine() any { return m.v }
// binding.Validator = &myValidator{v: validator.New(validator.WithRequiredStructEnabled())}
```

That's a lot of copied boilerplate for a common need.

**Proposal**

Provide a first-class way to construct the default validator with options. Any of:

1. An exported constructor for the default validator, e.g. `binding.NewStructValidator(opts ...validator.Option) StructValidator`, so users can do `binding.Validator = binding.NewStructValidator(validator.WithRequiredStructEnabled())` without re-implementing `ValidateStruct`.
2. A package-level setter such as `binding.SetValidatorOptions(opts ...validator.Option)` applied on `lazyinit`.
3. Exporting `defaultValidator` (or accepting options on it) so the ptr/slice dispatch can be reused.

Option 1 seems the least invasive and keeps gin's existing `ValidateStruct` behavior intact.

Tested with gin v1.12.0 and go-playground/validator v10.30.2.

sonnemusk:
Fixes #4733

`validator.WithRequiredStructEnabled()` (and other constructor-only options) can't be applied through the current default validator — `lazyinit` always calls `validator.New()` with no args, and `Engine()` is already too late.

Adds `binding.NewStructValidator(opts ...validator.Option)` so you can do:

```go
binding.Validator = binding.NewStructValidator(validator.WithRequiredStructEnabled())
```

without reimplementing `ValidateStruct`. Default `binding.Validator` behavior is unchanged.

Test covers the required-on-struct-field case from the issue.

### DIFF
--- a/binding/default_validator.go
+++ b/binding/default_validator.go
@@ -16,6 +16,18 @@ import (
 type defaultValidator struct {
 	once     sync.Once
 	validate *validator.Validate
+	opts     []validator.Option
+}
+
+// NewStructValidator returns a StructValidator backed by go-playground/validator
+// with the given options applied at construction time.
+//
+// This is needed for options that can only be set via validator.New, such as
+// validator.WithRequiredStructEnabled(). Example:
+//
+//	binding.Validator = binding.NewStructValidator(validator.WithRequiredStructEnabled())
+func NewStructValidator(opts ...validator.Option) StructValidator {
+	return &defaultValidator{opts: opts}
 }
 
 type SliceValidationError []error
@@ -92,7 +104,7 @@ func (v *defaultValidator) Engine() any {
 
 func (v *defaultValidator) lazyinit() {
 	v.once.Do(func() {
-		v.validate = validator.New()
+		v.validate = validator.New(v.opts...)
 		v.validate.SetTagName("binding")
 	})
 }
--- a/binding/default_validator_test.go
+++ b/binding/default_validator_test.go
@@ -7,6 +7,8 @@ package binding
 import (
 	"errors"
 	"testing"
+
+	"github.com/go-playground/validator/v10"
 )
 
 func TestSliceValidationError(t *testing.T) {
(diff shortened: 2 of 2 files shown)


sonnemusk:

Issue #3443, opened 2022-12-20
Status() not setting up status in response for unit test


infected-engineer:
## Description
ctx.Status(http.StatusCreated) is not setting up status in response and I am always getting 200.
But If I add `ctx.Writer.WriteHeaderNow()` with ctx.Status() then it is working fine.

## How to reproduce
```
// Sample Function
func PostCategory(ctx *gin.Context) {
	ctx.Status(http.StatusCreated)
}

// Sample Test Function
func TestPostCategory(t *testing.T) {
	r := httptest.NewRequest(http.MethodGet, "/", nil)
	w := httptest.NewRecorder()
	ctx, _ := gin.CreateTestContext(w)
	ctx.Request = r

	PostCategory(ctx)
	assert.Equal(t, http.StatusCreated, w.Code)
}
```

## Expectations
We must get 201 status code and test must be passed.

## Actual result
We are getting 2022 status code in response and test are failing

## Environment
- go version:  go1.18.8 linux/amd64
- gin version: v1.8.1
- operating system: Ubuntu 18.04 LTS

webstradev:
I'm not sure this is the answer that you are looking for, but it looks like you aren't actually serving the request.

When unit testing gin handlers you will likely want to actually register the route on a gin engine and serve the request.

Consider using the `gin.Engine` that is returned from the `gin.CreateTextContext(w)` function.

Your code would then look something like this

``` go
// Sample Function
func PostCategory(ctx *gin.Context) {
	ctx.Status(http.StatusCreated)
}

// Sample Test Function
func TestPostCategory(t *testing.T) {
	r := httptest.NewRequest(http.MethodGet, "/", nil)
	w := httptest.NewRecorder()
	ctx, router := gin.CreateTestContext(w) // I've changed this line to use the gin engine

	router.GET("/", PostCategory) // I've added this line to register a route using the function you want to test

	router.ServeHTTP(w, r) // I've added this line to actually serve the request to the router we've defined above

	PostCategory(ctx)

	assert.Equal(t, http.StatusCreated, w.Code)
}
```

Let me know if that helps 😄

infected-engineer:
I'm still considering it as bug because if you execute the given example in issue's description with standard library without using gin then you will status set in w.Code.

The reason is because gin is not setting up status in w.Code while using `httptest.NewRecorder()` 
[...]

Issue #3137, opened 2022-05-05
gin/context Copy method comment is confuse 


fengbenming:
- With issues:
  - gin/context Copy method comment is confuse. It'll be invalid memory address or nil pointer dereference when context copyed be passed to a goroutine.

## Description

// Copy returns a copy of the current context that can be safely used outside the request's scope.
// This has to be used when the context has to be passed to a goroutine.

It is not outside the request's scope, a copy of the current context just in the request's scop.

## How to reproduce

```
func CreateVulJob(c *gin.Context) {
	env := c.Param("env")
	if env == "aliyun" {
		go doSomething(c.Copy())
	} else{
	       go doOtherSomething(c.Copy())
	}
}
```

## Expectations

```
not panic
```

## Actual result

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x931ae4]

goroutine 100 [running]:
github.com/gin-gonic/gin.(*responseWriter).Header(0x30?)
        <autogenerated>:1 +0x24
github.com/gin-gonic/gin/render.writeContentType({0x7fe3541d1f58?, 0xc0003c5300?}, {0x24d3b40, 0x1, 0x1})
        /go/pkg/mod/github.com/gin-gonic/gin@v1.7.7/render/render.go:36 +0x3a
github.com/gin-gonic/gin/render.WriteJSON({0x7fe3541d1f58, 0xc0003c5300}, {0xd0ae40, 0xc0002865f0})
        /go/pkg/mod/github.com/gin-gonic/gin@v1.7.7/render/json.go:68 +0x4c
github.com/gin-gonic/gin/render.JSON.Render(...)
        /go/pkg/mod/github.com/gin-gonic/gin@v1.7.7/render/json.go:55
github.com/gin-gonic/gin.(*Context).Render(0xc0003c5300, 0x190, {0xebec20, 0xc0004241b0})
        /go/pkg/mod/github.com/gin-gonic/gin@v1.7.7/context.go:927 +0xf8
github.com/gin-gonic/gin.(*Context).JSON(...)
        /go/pkg/mod/github.com/gin-gonic/gin@v1.7.7/context.go:970
```

## Environment

- go version: go1.18.1 linux/amd64
- gin version (or commit ref):
- operating system: centos 7.0

capric8416:
**Can't reproduce using the following code, please paste more code snippet**
```golang
package main

import (
        "fmt"

[...]

PR #4742, opened 2026-07-21
docs(context): clarify Copy and Status behavior

sonnemusk:
### Description

Clarify two commonly confusing Context APIs:

1. **Copy** — still for use outside the handler (e.g. goroutines), but documents that the copy’s Writer is detached (`ResponseWriter == nil`) so `JSON` / other write methods panic. Addresses the confusion in #3137.
2. **Status** — documents that the code is buffered until flush/`WriteHeaderNow`, which is why unit tests that only call `Status` and assert `httptest.ResponseRecorder.Code` see 200. Related to #3443.

No runtime behavior change.

### Checklist
- [x] Documentation only
- [x] Related issues: #3137, #3443

### DIFF
--- a/context.go
+++ b/context.go
@@ -117,8 +117,13 @@ func (c *Context) reset() {
 	*c.skippedNodes = (*c.skippedNodes)[:0]
 }
 
-// Copy returns a copy of the current context that can be safely used outside the request's scope.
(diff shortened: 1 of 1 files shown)


sonnemusk:
Closing — pure docs/typo cleanup. Going to focus on behavioral fixes instead.

PR #4745, opened 2026-07-21
fix(context): encode FileAttachment filename* with PathEscape

sonnemusk:
## Summary

For non-ASCII download names, `FileAttachment` builds a RFC 5987 `filename*` parameter with `url.QueryEscape`. That encoder is for form queries and turns spaces into `+`, so a name like `报告 测试.pdf` becomes `报告+测试.pdf` in many clients.

Switch to `url.PathEscape` so spaces become `%20` (and the encoding matches percent-encoded attr-char expectations).

## Test plan

- [x] Updated UTF-8 attachment test to expect `PathEscape`
- [x] Added regression test for non-ASCII filename containing a space
- [x] `go test -count=1 -run 'TestContextRender.*Attachment' .`
- [x] `go test -count=1 .`

### DIFF
--- a/context.go
+++ b/context.go
@@ -1360,7 +1360,9 @@ func (c *Context) FileAttachment(filepath, filename string) {
 	if isASCII(filename) {
 		c.Writer.Header().Set("Content-Disposition", `attachment; filename="`+escapeQuotes(filename)+`"`)
 	} else {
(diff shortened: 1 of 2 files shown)


sonnemusk:
Closing: lower priority / not a must-have fix. Keeping a smaller set of high-impact PRs only. Sorry for the noise.

PR #4749, opened 2026-07-21
fix(binding): do not panic ValidateStruct on nil pointers

sonnemusk:
Nil pointers hit Elem().Interface() and panic despite the never-panic contract.


sonnemusk:
Closing this — it was a defensive nil guard, not something I hit in real use. Callers should not pass nil here; panicking (or failing at the call site) is fine.

PR #4750, opened 2026-07-21
fix: make nameOfFunction nil-safe for HandlerName

sonnemusk:
Copy()/empty chains yield nil handlers; FuncForPC(0).Name() panics.


sonnemusk:
Closing this — it was a defensive nil guard, not something I hit in real use. Callers should not pass nil here; panicking (or failing at the call site) is fine.

PR #4752, opened 2026-07-21
fix(binding): run validator after BSON unmarshal

sonnemusk:
BSON binding skipped validate(); JSON/YAML/XML all run it.


sonnemusk:

PR #4754, opened 2026-07-21
feat: add RFC 9457 problem details JSON response support

miqui:
## Summary

Adds first-class support for [RFC 9457 — Problem Details for HTTP APIs](https://www.rfc-editor.org/rfc/rfc9457) error responses in JSON format, as a purely additive API (no change to any existing default behavior):

- **`gin.Problem`** — struct with the standard members (`type`, `title`, `status`, `detail`, `instance`) and an `Extensions map[string]any` serialized as top-level members of the problem details object (standard members take precedence on key collisions). An empty `Type` is omitted, which RFC 9457 §3.1.1 defines as equivalent to `about:blank`.
- **`render.ProblemJSON`** — renderer setting `Content-Type: application/problem+json; charset=utf-8`.
- **`Context.ProblemJSON(code, obj)`** / **`Context.AbortWithProblemJSON(code, obj)`** — mirror the existing `JSON` / `AbortWithStatusJSON` conventions. When `obj` is a `Problem` or `*Problem`, a zero `Status` defaults to the response code and an empty `Title` defaults to `http.StatusText` (per the RFC's guidance for `about:blank` problems), without mutating the caller's value.
- **`gin.ProblemDetails()`** middleware — renders errors attached via `Context.Error` as a problem details response when the handler chain finishes without writing one; keeps the status when it is already an error code, otherwise defaults to 500.

## Example

```go
c.ProblemJSON(http.StatusForbidden, gin.Problem{
    Type:       "https://example.com/probs/out-of-credit",
    Detail:     "Your current balance is 30, but that costs 50.",
    Extensions: map[string]any{"balance": 30},
})
// HTTP 403, Content-Type: application/problem+json; charset=utf-8
// {"type":"https://example.com/probs/out-of-credit","title":"Forbidden","status":403,
//  "detail":"Your current balance is 30, but that costs 50.","balance":30}
```

## Pull Request Checklist

- [x] Tests are added or modified as needed to cover code changes (13 new tests, also verified with the `sonic` build tag).


miqui:
Long time user of Gin.  Thought would be a good idea to support RFC9457 http problem.

PR #4768, opened 2026-07-29
fix: add default ReadHeaderTimeout to http.Server in Run()

### LINKED ISSUE #4760: Engine.Run starts an HTTP server without default timeouts
## Summary

`Engine.Run` starts an `http.Server` without read, write, idle, or header timeouts.

This leaves applications that use the default `Run` helper exposed to slow-client resource exhaustion unless they manually create and configure their own `http.Server`.

## Affected Code

`gin.go`

```go
server := &http.Server{ // #nosec G112
    Addr:    address,
    Handler: engine.Handler(),
}
err = server.ListenAndServe()
```

`ReadTimeout`, `WriteTimeout`, `ReadHeaderTimeout`, and `IdleTimeout` are all left at their zero values.

## Reproduction

I reproduced this with a minimal Gin server started through the default run path.

A TCP client sent only a partial HTTP request header, waited 20 seconds, and then completed the request:

```text
GET /ping HTTP/1.1
Host: localhost
```

After the 20 second pause, the connection was still alive and the server returned a normal response once the request was completed:

```text
HTTP/1.1 200 OK
Content-Type:
```

This shows that no `ReadHeaderTimeout` was applied by default.

## Expected Behavior

The default server startup helper should either apply safe timeout defaults or strongly guide users toward a timeout-configured server for production use.

## Actual Behavior

`Engine.Run` starts an HTTP server with no default timeouts.

## Impact

Slow clients can keep connections open by delaying request headers. With enough concurrent slow connections, this can consume server connection resources and degrade availability.

## Suggested Fix

- Consider setting a conservative default `ReadHeaderTimeout` in the `Run` helper.
- Alternatively, add a production-safe helper or prominent documentation warning that `Engine.Run` does not configure timeouts.
- Recommend using an explicitly configured `http.Server` for production deployments.

### Gin Version

Local gin-gonic/gin checkout at commit 34dac20

Yes

### Source Code

```go
package main

import (
	"net/http"

	"github.com/gin-gonic/gin"
)

func main() {
	r := gin.Default()

	})

	if err := r.Run(":8080"); err != nil {
		panic(err)
	}
}
```

Run the server:

```bash
go run .
```

In another shell, open a TCP connection and send only a partial HTTP request header:

```bash
exec 3<>/dev/tcp/localhost/8080
printf 'GET /ping HTTP/1.1\r\nHost: localhost\r\n' >&3
sleep 20
printf '\r\n' >&3
timeout 3 dd bs=1 count=80 <&3 2>/dev/null
```

Observed behavior:

```text
HTTP/1.1 200 OK
Content-Type:
```

The connection remains open after 20 seconds and the server responds normally once the request header is completed, showing that `ReadHeaderTimeout` is not applied by default.

### Go Version

Go 1.26.5 in Docker golang:1.26

### Operating System

Linux, inside Docker golang:1.26 container

waterWang:
## Description

`Engine.Run` creates an `http.Server` without a `ReadHeaderTimeout`, making it susceptible to slow‑client resource exhaustion (slow‑loris style attacks). A client that sends only a partial HTTP request header can hold the connection open indefinitely.

## Fix

Added a default `ReadHeaderTimeout` of 10 seconds to the `http.Server` created by `Engine.Run()`. This is a safe, conservative default that prevents slow‑header attacks without breaking existing applications.

The `#nosec G112` comment on the struct literal is kept, because the G112 linter flags the *absence* of `ReadHeaderTimeout`, `ReadTimeout`, and `WriteTimeout` — and we are intentionally only setting `ReadHeaderTimeout` here. `ReadTimeout` and `WriteTimeout` are not set because they would break long‑running request handlers (e.g. SSE, WebSocket, large uploads), which is a deliberately different design concern.

## Related Issue

Fixes #4760

## Checklist

- [x] Code compiles (`go build ./...`)
- [x] Tests pass (`go test ./...`)
- [x] `gofmt` is clean

### DIFF
--- a/gin.go
+++ b/gin.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"time"
 	"path"
 	"strings"
 	"sync"
@@ -548,8 +549,9 @@ func (engine *Engine) Run(addr ...string) (err error) {
 	address := resolveAddress(addr)
 	debugPrint("Listening and serving HTTP on %s\n", address)
 	server := &http.Server{ // #nosec G112
-		Addr:    address,
-		Handler: engine.Handler(),
+		Addr:              address,
+		Handler:           engine.Handler(),
+		ReadHeaderTimeout: 10 * time.Second,
 	}
 	err = server.ListenAndServe()
 	return


gterminator:
copy of #4764 ?

waterWang:
Hi maintainers, just a friendly ping — this PR has been open for a while. Could you take a look when you have a moment? Happy to address any feedback. Thanks!

PR #4783, opened 2026-08-02
fix(tree): roll back to skipped wildcards at param route dead ends

SetagGnaw:
# Pull Request Checklist

- [ ] If the pull request introduces a new feature, the feature is documented in the `docs/doc.md`. (bug fix, no new feature)

## Description

This PR fixes two related correctness bugs in the router's wildcard backtracking (the `skippedNodes` mechanism).

### 1. Param route dead ends never fall back to skipped wildcards

`getValue` pushes a `skippedNode` entry whenever it prefers a static child over a wildcard sibling, and rolls back to it when the static branch fails. Two dead ends inside the `param` case were missing that rollback and returned "not found" directly:

- more path remains but the param node has no child to descend into
- the param node consumed the whole remaining path but has no handler registered

Reproduction:

```go
router.GET("/:p0/:p1", handler)
router.GET("/bc/:p1/a", handler)
// GET /bc/bc returned 404; it now matches /:p0/:p1
```

The walk prefers the static `bc` branch, dead-ends on the handler-less `:p1` node, and previously gave up instead of retrying the skipped `/:p0` branch.

Both dead ends now roll back to the last valid `skippedNode`, consistent with the existing fallback at the other dead ends of `getValue`. Since this would have added a fourth and fifth copy of the same inline rollback loop, the loop is extracted into a `rollbackToSkippedNode` helper used by all five sites (net behavior of the three existing sites is unchanged).

### 2. Stale skippedNodes corrupt 405 Method Not Allowed probes

`handleHTTPRequest` reuses `c.skippedNodes` across the per-method `getValue` probes that build the `Allow` header when `HandleMethodNotAllowed` is enabled. A probe that matches via a static branch leaves its skipped wildcard entries on the stack; when a later method tree's probe dead-ends, it can roll back through such a stale entry into the previous tree's nodes, find a handler there, and wrongly report its own method as allowed. With the routes in the new test, `GET /b` answered `Allow: POST, PUT` instead of `Allow: POST`. The stack is now reset before each probe.

### Tests

- `TestTreeParamFallbackAfterStaticDeadEnd` covers each of the two dead ends (each subtest fails without its corresponding fix).
- `TestRouteNotAllowedDoesNotReuseSkippedNodes` covers the 405 probe reset.
- `go test -race ./...` passes and `golangci-lint run` reports 0 issues.

### DIFF
--- a/gin.go
+++ b/gin.go
@@ -743,6 +743,7 @@ func (engine *Engine) handleHTTPRequest(c *Context) {
 			if tree.method == httpMethod {
 				continue
 			}
+			*c.skippedNodes = (*c.skippedNodes)[:0]
 			if value := tree.root.getValue(rPath, nil, c.skippedNodes, unescape); value.handlers != nil {
 				allowed = append(allowed, tree.method)
 			}
--- a/routes_test.go
+++ b/routes_test.go
@@ -527,6 +527,23 @@ func TestRouteNotAllowedEnabled3(t *testing.T) {
 	assert.Contains(t, allowed, http.MethodPost)
 }
 
+func TestRouteNotAllowedDoesNotReuseSkippedNodes(t *testing.T) {
+	router := New()
+	router.HandleMethodNotAllowed = true
+	router.POST("/b", func(c *Context) {})
+	router.POST("/:p0", func(c *Context) {})
+	router.PUT("/a/:p1", func(c *Context) {})
+
+	// The POST probe for the Allow header matches static "/b" while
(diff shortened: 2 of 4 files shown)


rashmi-tondare:
This PR also fixes #4818 -- the same stale `c.skippedNodes` reuse panics `getValue` outright, not just corrupts the `Allow` header. The one-line reset in `gin.go` here is sufficient, I confirmed it against v1.12.0.

PR #4784, opened 2026-08-04
feat(binding): support validator constructor options

### LINKED ISSUE #4733: Allow configuring the binding validator with go-playground/validator Options (e.g. WithRequiredStructEnabled)

stareezy-1:
## Description

Gin owns reusable struct, pointer, slice, and array validation dispatch, but the default implementation hardcodes `validator.New()` inside its private lazy initializer. This prevents callers from applying constructor-only validator options without reimplementing `StructValidator`.

This adds `binding.NewStructValidator(options...)`, passes a private copy of those options into the existing lazy initializer, adds regression coverage for `WithRequiredStructEnabled`, and documents configuration before request handling.

Fixes #4733

## Verification

- `go test ./binding -count=1`
- `go test -tags nomsgpack ./binding -count=1`
- `go test -race ./binding -run Test(NewStructValidatorWithOptions|DefaultValidator) -count=1`
- `make test`
- `make vet`

# Pull Request Checklist

- [x] The new feature is documented in `docs/doc.md`.

### DIFF
--- a/binding/default_validator.go
+++ b/binding/default_validator.go
@@ -16,6 +16,15 @@ import (
 type defaultValidator struct {
 	once     sync.Once
 	validate *validator.Validate
+	options  []validator.Option
+}
+
+// NewStructValidator returns Gin's default StructValidator configured with
+// options passed to validator.New.
+func NewStructValidator(options ...validator.Option) StructValidator {
+	return &defaultValidator{
+		options: append([]validator.Option(nil), options...),
+	}
 }
 
 type SliceValidationError []error
@@ -92,7 +101,7 @@ func (v *defaultValidator) Engine() any {
 
 func (v *defaultValidator) lazyinit() {
 	v.once.Do(func() {
-		v.validate = validator.New()
+		v.validate = validator.New(v.options...)
 		v.validate.SetTagName("binding")
 	})
 }
--- a/binding/default_validator_test.go
+++ b/binding/default_validator_test.go
@@ -7,6 +7,8 @@ package binding
 import (
 	"errors"
 	"testing"
+
+	"github.com/go-playground/validator/v10"
 )
 
 func TestSliceValidationError(t *testing.T) {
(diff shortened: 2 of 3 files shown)


stareezy-1:
CI note: lint, CodeQL, and the complete Go 1.25/1.26 test matrix pass. The failing Trivy job reports CVE-2026-56852 in the `golang.org/x/text v0.37.0` version already pinned on upstream `master` (fixed in v0.39.0); this PR does not modify `go.mod`, so I kept that repository-wide dependency upgrade out of this focused change.

PR #4793, opened 2026-08-08
fix(binding): include field name in BindUnmarshaler errors

### LINKED ISSUE #4296: BindUnmarshaler should include the field name in error messages
Hi, little issue here when writing custom unmarshalling interfaces.

I have a simple interface that binds UUID types. I noticed that when an error occurs during parsing, the error message in the response just includes the raw error text and no mention of the field name. As a user, receiving an error text without the actual field name can be very confusing.

There's no access to the field name in the UnmarshalParam implementation, and it seems quite nontrivial to otherwise insert the field name that is erroring. We could use a level of convenience on top of this, either automatically including the field name in the error response, or letting us have easier access to it during unmarshalling.

RubenPari:
## Summary
- Prefix `BindUnmarshaler` / `encoding.TextUnmarshaler` bind errors with the form/query tag name
- Nested element path uses the Go field name (no parent tag in scope)
- Errors stay unwrap-friendly via `%w`

## Test plan
- [x] `go test ./binding -count=1 -run 'TestBindUnmarshalerErrorIncludesFieldName|TestMappingCustomUnmarshal'`

Fixes #4296

### DIFF
--- a/binding/form_mapping.go
+++ b/binding/form_mapping.go
@@ -196,6 +196,14 @@ func trySetCustom(val string, value reflect.Value) (isSet bool, err error) {
 	return false, nil
 }
 
+// wrapBindErr prefixes a bind error with the form/query field name.
+func wrapBindErr(fieldName string, err error) error {
+	if err == nil || fieldName == "" {
+		return err
+	}
+	return fmt.Errorf("%s: %w", fieldName, err)
+}
+
 // trySetUsingParser tries to set a custom type value based on the presence of the "parser" tag on the field.
 // If the parser tag does not exist or does not match any of the supported parsers, gin will skip over this.
 func trySetUsingParser(val string, value reflect.Value, parser string) (isSet bool, err error) {
@@ -264,9 +272,9 @@ func setByForm(value reflect.Value, field reflect.StructField, form map[string][
 		}
 
 		if ok, err = trySetUsingParser(vs[0], value, opt.parser); ok {
-			return ok, err
+			return ok, wrapBindErr(tagValue, err)
 		} else if ok, err = trySetCustom(vs[0], value); ok {
-			return ok, err
+			return ok, wrapBindErr(tagValue, err)
 		}
 
 		if vs, err = trySplit(vs, field); err != nil {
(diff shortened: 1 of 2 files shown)


wahajahmed010 (commented):
Nice fix — adding the field tag/name to `UnmarshalParam` / `TextUnmarshaler` errors is a real usability win. A few findings:

**1. Inconsistent field naming between `setByForm` and `setWithProperType`.**

- `setByForm` wraps with `tagValue` (the `form:"xxx"` / `query:"xxx"` tag value).
- `setWithProperType` wraps with `field.Name` (the Go field name).

The PR description says this is intentional because `setWithProperType` does not have a tag in scope — but it actually does: `field reflect.StructField` has both `Name` and `Tag`, and `opt.tag` (set earlier) holds the parsed tag value. See line ~245 where `tagValue` is populated. So both call sites have access to the tag, but the new code uses different sources. Recommend standardizing on the tag value for both, since that is what the user wrote in their struct definition and what `c.Bind()` will return.

If the asymmetry is intentional (e.g. for nested fields where the tag is empty), please add a comment to `wrapBindErr` or `setWithProperType` explaining when each branch fires.

**2. Test coverage is incomplete.** The new `TestBindUnmarshalerErrorIncludesFieldName` only covers the `setByForm` path via `mappingByPtr`. Missing:

- A test for the `setWithProperType` path (where `field.Name` is the wrapper source).
- A test with `query:"..."` tag to confirm both code paths.
- A test with a nested struct field that has a `form` tag.
- A test where `UnmarshalParam` returns `nil` (should still pass through unchanged, since `wrapBindErr` short-circuits).

**3. `trySetUsingParser` error path:** if `parser` is empty (parser tag absent), `trySetUsingParser` returns `(false, nil)`, so `ok` is false and `wrapBindErr` is never reached. Good — no change needed, but worth a test asserting no spurious wrapping when no parser is configured.

**4. `fmt.Errorf("%s: %w", fieldName, err)` is safe for normal tag values.** Tag values cannot contain `%` in valid Go (it would be a parse error), so the format-string vulnerability does not apply. The `errors.New("boom")` in the test confirms unwrapping works.

**5. Consider renaming `wrapBindErr` → `wrapFieldBindErr` or `prefixBindErr` for clarity.** Currently it reads as if it wraps a binder, not wraps an error with a field name.

Overall: the fix is correct, but the tag-vs-name inconsistency is the biggest concern — it will lead to confusing behavior where the same struct field produces `"uuid4: boom"` from one path and `"ID: boom"` from another. Recommend resolving before merge.

PR #4796, opened 2026-08-08
feat(engine): optional skip of global middleware on 405

### LINKED ISSUE #4189: Engine.NoMethod() is ignored for middlewares
when Engine.NoMethod() is used with some handler, where is no way to restrict trigger middlewares, passed with Engine.Use(), its confusing, because if i have, for example 
handler
`
    func NoMethodHandler(c *gin.Context) {
        c.String(http.StatusMethodNotAllowed, "method not allowed")
        c.Abort()
    }
` 

and some middleware, works fine only for POST requests:
`
    func CheckSummMiddleware(c *gin.Context) {
        //read req body and "X-Checksum" header
        if !checkSumm(bodybytes, headerValue){
             c.String(http.StatusBadRequest, "wrong checksum")
             c.Abort()
        }
    }
` 

after starting engine with 
`
    handler := gin.New()
    handler.NoRoute(NoMethodHandler)
    handler.Use(CheckSummMiddleware)
    hendler.POST("/ping",func(c *gin.Context){
    } )
    handler.Run(":{{$PORT}}")
`
i expect what if ill `curl --location '{{$ADDRESS}}:{{PORT}}/ping'`  ill get 
"405 method not allowed", but actually result is "400 wrong checksum"

i think where should be an public variable in Engine like 
`SkipAllOnMethodNotAllowed bool`

and some changes in 
`func (engine *Engine) rebuild405Handlers()`  (https://github.com/gin-gonic/gin/blob/ebe5e2a6bfdca50fd44074b470ad486392e2933f/gin.go#L337)
like 
`
...
if engine.SkipAllOnMethodNotAllowed {
    engine.allNoMethod = engine.noMethod
        return
    }
...
`

RubenPari:
## Summary
- New opt-in `Engine.NoMethodSkipHandlers` (default `false`)
- When true (and `HandleMethodNotAllowed` is on), the 405 chain is only `NoMethod` handlers — global `Use` middleware is not prepended
- Lets body-reading / aborting middleware avoid turning GET-on-POST into 400 instead of 405
- Set the flag before `Use` / `NoMethod` (rebuild runs on those calls)

## Test plan
- [x] `go test . -count=1 -run 'TestNoMethod|TestMiddlewareNoMethod'`

Fixes #4189

### DIFF
--- a/gin.go
+++ b/gin.go
@@ -122,6 +122,13 @@ type Engine struct {
 	// handler.
 	HandleMethodNotAllowed bool
 
+	// NoMethodSkipHandlers when true, the 405 chain is only Engine.NoMethod
+	// handlers — global middleware from Engine.Use is not prepended. Default
+	// false keeps historical behavior (logger/recovery still run on 405).
+	// Only applies when HandleMethodNotAllowed is true.
+	// Set this before Use/NoMethod (rebuild happens on those calls).
+	NoMethodSkipHandlers bool
+
 	// ForwardedByClientIP if enabled, client IP will be parsed from the request's headers that
 	// match those stored at `(*gin.Engine).RemoteIPHeaders`. If no IP was
 	// fetched, it falls back to the IP obtained from
@@ -358,6 +365,10 @@ func (engine *Engine) rebuild404Handlers() {
 }
 
 func (engine *Engine) rebuild405Handlers() {
+	if engine.NoMethodSkipHandlers {
+		engine.allNoMethod = engine.noMethod
+		return
+	}
 	engine.allNoMethod = engine.combineHandlers(engine.noMethod)
 }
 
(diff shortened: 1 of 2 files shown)


BitWeaverDev:
Hi @RubenPari, sorry for the duplicate work. I opened #4804 for the same issue before I spotted this PR. My search for existing PRs on #4189 came back stale and missed it, so yours came first by two days. I've said on mine that I'll close it if the maintainers prefer this one.

Since I ended up writing the same option, here is one thing that may be useful to you either way.

Deciding inside `rebuild405Handlers()` makes the flag sensitive to registration order, which is why you needed the caveat in the field doc ("Set this before Use/NoMethod"). The caveat is accurate, but the failure mode is silent. This compiles, reads as correct, and does nothing:

```go
r := gin.New()
r.HandleMethodNotAllowed = true
r.Use(auth)
r.NoMethod(h)
r.NoMethodSkipHandlers = true   // no rebuild is triggered; allNoMethod still contains auth
```

`rebuild405Handlers()` only runs from `Use()` and `NoMethod()`. Assigning to a plain struct field triggers nothing, so the chain keeps the middleware. Both tests here set the flag before `Use`/`NoMethod`, so this path is not covered at the moment.

If you resolve the flag when the request is served rather than when the chain is built, the caveat goes away:

```go
func (engine *Engine) noMethodHandlers() HandlersChain {
	if engine.NoMethodSkipHandlers {
		return engine.noMethod
	}
	return engine.allNoMethod
}
```

and in `handleHTTPRequest`, `c.handlers = engine.noMethodHandlers()`. `rebuild405Handlers()` stays as it is, so `allNoMethod` keeps its current meaning and no existing test needs to change. It also costs nothing on the hot path, since the branch only sits on the 405 error path.

The PR checklist also asks for new features to be documented in `docs/doc.md`, which is not in this diff yet.

Take any of this if it helps.

PR #4798, opened 2026-08-08
fix(context): prevent nil pointer panic in PostForm getters when Request is nil

### LINKED ISSUE #4772: PostForm getters panic when Context has no request
### Description

`Context.GetPostForm`, `Context.PostForm`, and `Context.DefaultPostForm` panic when `c.Request` is nil. This differs from the equivalent query getters, which return empty values safely.

### Reproduction

```go
c, _ := gin.CreateTestContext(httptest.NewRecorder())
c.PostForm("key") // panic: nil pointer dereference
```

### Expected behavior

The form getters should behave like the query getters for an empty context:

- `GetPostForm("key")` returns `("", false)`
- `PostForm("key")` returns `""`
- `DefaultPostForm("key", "fallback")` returns `"fallback"`

### Actual behavior

`initFormCache` calls `c.Request.ParseMultipartForm(...)` without checking whether `c.Request` is nil, causing a nil-pointer panic.

### Proposed fix

Initialize an empty form cache and return when there is no request. This preserves normal request behavior and matches `initQueryCache`.

### Validation

A regression test covering all three getters passes, along with `go test ./...` and `go vet ./...`.

### Environment

- Gin: master at `34dac20`
- Go: 1.26.4
- OS: macOS

RubenPari:
## Summary

`Context.GetPostForm`, `Context.PostForm`, `Context.DefaultPostForm`, and the array/map variants panic when `c.Request` is nil. This differs from the equivalent query getters, which return empty values safely.

### Root cause

`initFormCache` called `c.Request.ParseMultipartForm(...)` without checking whether `c.Request` is nil, causing a nil-pointer dereference.

### Fix

Added a nil guard in `initFormCache` — when `c.Request` is nil, an empty `url.Values` cache is initialized and returned, matching the behavior of `initQueryCache`.

### Behavior after fix

- `GetPostForm("key")` returns `("", false)`
- `PostForm("key")` returns `""`
- `DefaultPostForm("key", "fallback")` returns `"fallback"`

## Test plan

- [x] `TestContextPostFormOnEmptyRequest` — verifies all PostForm getters (including array and map variants) do not panic and return empty/default values when `c.Request == nil`
- [x] `go test . -race` — all existing tests pass

Fixes #4772

### DIFF
--- a/context.go
+++ b/context.go
@@ -649,12 +649,14 @@ func (c *Context) initFormCache() {
 	if c.formCache == nil {
 		c.formCache = make(url.Values)
 		req := c.Request
-		if err := req.ParseMultipartForm(c.engine.MaxMultipartMemory); err != nil {
-			if !errors.Is(err, http.ErrNotMultipart) {
-				debugPrint("error on parse multipart form array: %v", err)
+		if req != nil {
+			if err := req.ParseMultipartForm(c.engine.MaxMultipartMemory); err != nil {
+				if !errors.Is(err, http.ErrNotMultipart) {
+					debugPrint("error on parse multipart form array: %v", err)
+				}
 			}
+			c.formCache = req.PostForm
 		}
-		c.formCache = req.PostForm
 	}
 }
 
--- a/context_test.go
+++ b/context_test.go
@@ -903,6 +903,31 @@ func TestContextDefaultQueryOnEmptyRequest(t *testing.T) {
 	})
 }
 
+func TestContextPostFormOnEmptyRequest(t *testing.T) {
+	c, _ := CreateTestContext(httptest.NewRecorder()) // here c.Request == nil
+	assert.NotPanics(t, func() {
+		value, ok := c.GetPostForm("NoKey")
+		assert.False(t, ok)
+		assert.Empty(t, value)
+	})
+	assert.NotPanics(t, func() {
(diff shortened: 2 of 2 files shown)


xhon-pelushi (approved):
Checked this against \`initQueryCache\` (context.go:578), which the description says this mirrors — it does: same nil-check-then-empty-\`url.Values{}\`-fallback shape, so this brings \`initFormCache\` in line with the existing idiom rather than inventing a new pattern.

Traced through the nil-\`Request\` path specifically: since the new \`req != nil\` guard wraps the whole block, \`c.engine.MaxMultipartMemory\` is never dereferenced when \`Request\` is nil either, so this doesn't trade a nil-\`Request\` panic for a nil-\`engine\` one in that path.

\`TestContextPostFormMalformedMultipart\` is a nice adjacent case to include — malformed multipart body with a multipart Content-Type takes a different branch than "no Request at all" (it hits the existing \`ParseMultipartForm\` error path, just now not crashing beforehand), so it's worth having its own regression test rather than relying on the nil-Request case alone. Confirmed \`req.PostForm\` is always non-nil by the time that assignment runs, per \`net/http\`'s \`ParseForm\`, so the debug-log-and-continue behavior on error was already safe and didn't need to change.

Small, correctly scoped, good test coverage. LGTM.

PR #4687, opened 2026-06-02
feat: add SkipMethodNotAllowedMiddleware option

goingforstudying-ctrl:
Add `Engine.SkipMethodNotAllowedMiddleware` flag that, when enabled, prevents global middleware from running on 405 Method Not Allowed responses. This allows NoMethod handlers to respond correctly without being intercepted by middleware such as authentication or checksum validation that would otherwise reject the request first.

Fixes #4189

Issue #4309, opened 2025-07-22
Test Coverage Failure: Go 1.23+ Warning Branch Not Covered in CI Pipeline


Caplost:
## Problem Description

The CI automated testing pipeline is failing due to insufficient test coverage in the `debug.go` file. Specifically:
- `debugPrint` function coverage: 75.0%
- `debugPrintWARNINGDefault` function coverage: 66.7%

## Root Cause Analysis

The issue stems from the Go version check logic in `debug.go` lines 79-83:

```go
func debugPrintWARNINGDefault() {
    if v, e := getMinVer(runtime.Version()); e == nil && v < ginSupportMinGoVer {
        debugPrint(`[WARNING] Now Gin requires Go 1.23+.

`)
    }
    debugPrint(`[WARNING] Creating an Engine instance with the Logger and Recovery middleware already attached.

`)
}
```

**Core Issue:**
- CI environment runs on Go 1.23+ version
- Condition `v < ginSupportMinGoVer` (23) evaluates to false
- The Go 1.23+ warning branch is never executed
- This code branch remains untested, causing coverage failure

## Impact Assessment

1. **CI Pipeline Failure**: Test coverage checks fail to pass
2. **Code Quality Risk**: Untested code branches may contain potential issues
3. **Development Workflow Blockage**: PRs cannot be merged normally
4. **Release Process Disruption**: Automated deployment may be affected

### Solution 1: Mock Low Go Version Environment
```go
func TestDebugPrintWARNINGDefaultLowGoVersion(t *testing.T) {
    // Temporarily modify ginSupportMinGoVer or use mocking
    // Ensure Go version warning branch is tested
    originalMinVer := ginSupportMinGoVer
    ginSupportMinGoVer = 25 // Force condition to be true
    defer func() { ginSupportMinGoVer = originalMinVer }()
    
    // Test the warning branch
    re := captureOutput(t, func() {
        SetMode(DebugMode)
        debugPrintWARNINGDefault()
        SetMode(TestMode)
    })
    assert.Contains(t, re, "[WARNING] Now Gin requires Go 1.23+")
}
```

### Solution 2: Refactor Test Logic
- Extract version checking logic into a testable independent function
- Use dependency injection or interface mocking for different version environments
- Make the version comparison logic more testable

### Solution 3: Adjust Coverage Strategy
- Exclude specific conditional branches in CI configuration
- Lower coverage requirement threshold for this specific case
- Add coverage exclusion comments for environment-dependent code

[...]

PR #4804, opened 2026-08-10
feat(engine): add SkipMethodNotAllowedMiddleware option

### LINKED ISSUE #4189: Engine.NoMethod() is ignored for middlewares

BitWeaverDev:
This PR was written in part with the assistance of generative AI.

Fixes #4189

## Problem

Middleware registered via `Use()` is prepended to the `NoMethod` handlers by `rebuild405Handlers`, so a global middleware that aborts the request answers before the `NoMethod` handler ever runs. A middleware that validates a header only some methods carry therefore turns every 405 into its own error response:

```go
r := gin.New()
r.HandleMethodNotAllowed = true
r.Use(checkSum)                  // aborts with 400 when X-Checksum is missing
r.NoMethod(methodNotAllowed)     // never reached
r.POST("/ping", pingHandler)
```

`GET /ping` returns `400 wrong checksum` instead of `405 method not allowed`.

## Solution

Add an opt-in `Engine.SkipMethodNotAllowedMiddleware` flag that runs only the `NoMethod` handlers for 405 responses. It is disabled by default, so existing behaviour is unchanged, and `Use()`'s documented "Even 404, 405, static files..." contract still holds unless you opt out.

The flag is resolved when the request is served rather than when the handlers chain is built:

```go
func (engine *Engine) noMethodHandlers() HandlersChain {
	if engine.SkipMethodNotAllowedMiddleware {
		return engine.noMethod
	}
	return engine.allNoMethod
}
```

This matters. Deciding inside `rebuild405Handlers()` looks more natural, but that function only runs from `Use()` and `NoMethod()` — setting a plain struct field triggers no rebuild, so `r.Use(mw); r.NoMethod(h); r.SkipMethodNotAllowedMiddleware = true` would silently do nothing. Resolving at dispatch time makes the flag independent of registration order, and leaves `allNoMethod` semantics untouched, so no existing test needed modification.

Gin already uses these exact semantics internally: `createStaticHandler` dispatches the raw `noRoute` chain rather than `allNoRoute` so global middleware is not re-run for static-file misses (routergroup.go). This change exposes the equivalent behaviour for 405 as an explicit option.

Scope is deliberately limited to 405. `rebuild404Handlers()` has the same property, but this issue is about `NoMethod` and I would rather keep the diff reviewable; happy to extend to `NoRoute` in a follow-up if you would like it.

## Tests

Four tests in `middleware_test.go` alongside `TestMiddlewareNoMethodEnabled`, following its `signature`-string style so they assert handler ordering rather than just a boolean:

- `TestMiddlewareNoMethodSkipped` — flag on: 405, `Allow: POST`, signature `EGHF` (vs `ACEGHFDB`).
- `TestMiddlewareNoMethodSkippedSetAfterRegistration` — flag set after `Use()`/`NoMethod()` still takes effect.
- `TestMiddlewareNoMethodSkippedWithoutNoMethodHandlers` — falls back to the default 405 body, global middleware not run.
- `TestMiddlewareNoMethodSkippedDoesNotAffectNoRoute` — 404 path keeps running global middleware, locking in the 405-only scope.

I mutation-tested these by neutering the flag check: three fail with the exact pre-fix values, confirming they actually guard the behaviour. All existing 404/405 tests pass unmodified. Verified with the full suite, `-race`, `go vet`, `gofmt`, and end to end against a real server running the issue's scenario.

## Note on prior art

This supersedes #4687, which implemented the same option and was closed by its author for lack of review, not because of any objection. That version branched inside `rebuild405Handlers()` and so was sensitive to registration order as described above; `TestMiddlewareNoMethodSkippedSetAfterRegistration` is the regression guard for that case.

---

### DIFF
--- a/docs/doc.md
+++ b/docs/doc.md
@@ -24,6 +24,7 @@
   - [Custom Recovery behavior](#custom-recovery-behavior)
   - [Using BasicAuth() middleware](#using-basicauth-middleware)
   - [Goroutines inside a middleware](#goroutines-inside-a-middleware)
+  - [Skip middleware for 405 responses](#skip-middleware-for-405-responses)
 - [Logging](#logging)
   - [How to write log file](#how-to-write-log-file)
   - [Custom Log Format](#custom-log-format)
@@ -604,6 +605,51 @@ func main() {
   r.Run(":8080")
 }
 ```
+
+### Skip middleware for 405 responses
+
+Middleware registered with `Use()` runs for every request, including the `405 Method Not
+Allowed` responses produced when `HandleMethodNotAllowed` is enabled. A middleware that
+rejects the request therefore aborts the chain before the `NoMethod()` handler can reply,
+and the client receives the middleware's response instead of a 405.
+
+Enable `SkipMethodNotAllowedMiddleware` to run only the `NoMethod()` handlers for those
+responses:
+
+```go
+func main() {
+  r := gin.New()
+  r.HandleMethodNotAllowed = true
+
+  // Without SkipMethodNotAllowedMiddleware this middleware answers "wrong checksum"
+  // with 400 for a GET request, because a GET carries no X-Checksum header.
+  r.SkipMethodNotAllowedMiddleware = true
+
+  r.Use(func(c *gin.Context) {
+    if c.GetHeader("X-Checksum") == "" {
+      c.String(http.StatusBadRequest, "wrong checksum")
+      c.Abort()
+      return
+    }
+    c.Next()
+  })
+
+  r.NoMethod(func(c *gin.Context) {
+    c.String(http.StatusMethodNotAllowed, "method not allowed")
+  })
+
+  r.POST("/ping", func(c *gin.Context) {
+    c.String(http.StatusOK, "pong")
+  })
+
+  // GET /ping now returns 405 "method not allowed" with an "Allow: POST" header.
+  r.Run(":8080")
+}
+```
+
+The option is disabled by default, so the existing behaviour is unchanged. It only
+applies to 405 responses; requests handled by `NoRoute()` still run the global
+middleware.
 
 ## Logging
 
--- a/gin.go
+++ b/gin.go
@@ -122,6 +122,14 @@ type Engine struct {
 	// handler.
 	HandleMethodNotAllowed bool
 
+	// SkipMethodNotAllowedMiddleware if enabled, global middleware registered via Use()
+	// is not executed when the router answers with 405 Method Not Allowed. Only the
+	// handlers registered with NoMethod run. This lets a NoMethod handler reply with
+	// 405 even when a global middleware (authentication, checksum validation, ...)
+	// would otherwise abort the request first.
+	// Requires HandleMethodNotAllowed to be enabled.
+	SkipMethodNotAllowedMiddleware bool
+
 	// ForwardedByClientIP if enabled, client IP will be parsed from the request's headers that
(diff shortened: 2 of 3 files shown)


BitWeaverDev:
Cross-referencing: #4796 addresses the same issue and was opened two days before this PR. I missed it when I checked for existing work, so apologies for the duplicate. Maintainers should take whichever is more useful. I will close this one if #4796 is preferred, and I have left the substantive findings there so they are not lost either way.

The differences, for anyone deciding between them:

Registration order. #4796 branches inside `rebuild405Handlers()`, so the flag only takes effect if it is set before `Use()`/`NoMethod()`, which is documented there as a caveat. This PR resolves the flag when the request is served, so order does not matter and no caveat is needed. `TestMiddlewareNoMethodSkippedSetAfterRegistration` covers that case.

Documentation. `docs/doc.md` is required by the PR checklist for new features and is included here.

Tests. Four here against two, using the `signature` string style of the surrounding `TestMiddlewareNoMethod*` tests so ordering is asserted rather than just the status code. There is also coverage for the `Allow` header, the default 405 body when no `NoMethod` handler is registered, and the 404 path being unaffected.

Naming. `SkipMethodNotAllowedMiddleware` here against `NoMethodSkipHandlers` there. I have no strong preference and can adopt either.

On CI: rebased onto current master, which picked up the `golang.org/x/text` bump in #4807, so the Trivy failure is resolved. 26 checks pass.

The one remaining failure is `codecov/project`, which I do not believe this PR can address. `codecov/patch` passes, and running `make test` on this branch and on `master` (dcaa429) gives an identical total of 98.8% statements, so the change is coverage neutral. `codecov.yml` sets `project.target: 99%`, which master itself does not currently meet, so the gate fails on every open PR I checked (#4796, #4799, #4800, #4801). The base commit codecov is comparing against is also still `3dc1cd6`, 289 commits behind master, which matches the indexing problem reported in #4309. Both look like repository level configuration rather than anything in this diff, but let me know if you would like me to do something about it.

PR #4732, opened 2026-07-05
feat: add support for HTTP QUERY method

Aum-Patel1234:
feat: add support for HTTP QUERY method
Fixes #4721

# Pull Request Checklist

Please ensure your pull request meets the following requirements:

- [x] Open your pull request against the `master` branch.
- [ ] All tests pass in available continuous integration systems (e.g., GitHub Actions).
- [x] Tests are added or modified as needed to cover code changes.
- [x] If the pull request introduces a new feature, the feature is documented in the `docs/doc.md`.

Thank you for contributing!

Issue #4721, opened 2026-06-29
Support HTTP Query Method with rfc10008


FarmerChillax:
### Feature Description

ref: https://www.rfc-editor.org/info/rfc10008/

phil-smith-lu:
+1 for this. Would we need it in the `net/http` package first? https://github.com/golang/go/issues/80058

Aum-Patel1234:
Since RFC 10008 standardizes `QUERY` as an HTTP request method, I don't think Gin needs to wait for `net/http` to expose a dedicated helper first.

Gin already routes based on `Request.Method`, and since Go doesn't restrict HTTP methods to a predefined set, supporting `QUERY` can be as simple as adding a convenience router method (similar to `GET`, `POST`, etc.) that internally registers the `"QUERY"` method.

This would also allow developers to experiment with the newly standardized method while broader ecosystem support (including `net/http`) evolves. If `net/http` later adds first-class constants or helpers, Gin can easily align with those without changing its public API.

Of course, the semantics of an HTTP method ultimately depend on the application. A developer can technically perform a state-changing operation even in response to a `GET` request, despite HTTP semantics saying otherwise. In the same spirit, the framework can provide support for the standardized `QUERY` method, while it's up to developers to use it according to the semantics defined in RFC 10008.

Below is an implementation of the `QUERY` method based on my understanding of RFC 10008. I'd appreciate any feedback if I've misunderstood any part of the specification.

Jia-ben00:
I've opened PR #4830 implementing this feature. It adds a `QUERY()` router method as a shortcut for `Handle("QUERY", path, handlers)`, following the same pattern as GET/POST/HEAD.

The QUERY method is defined by RFC 10008. Since Go's net/http doesn't yet expose http.MethodQuery (golang/go#80058), the literal string "QUERY" is used, which works because Gin routes based on Request.Method and Go doesn't restrict HTTP methods to a predefined set.

PR #4806, opened 2026-08-14
feat: add support for the HTTP QUERY method (RFC 10008)

GP-09:
Ref #4721

## Pull Request Checklist

## What this adds

- `gin.MethodQuery` constant (`"QUERY"`).
- `RouterGroup.QUERY(path, handlers...)`, so `router.QUERY(...)` and `group.QUERY(...)` work like every other method shortcut.
- `ginS.QUERY(...)` for the package-level API.
- `gin.IQueryRoutes`, a one-method interface for callers that only hold an `IRoutes` value.

## Relation to #4732

#4732 proposes the same feature and stalled on the review question raised there: how to add `QUERY` without degenerating the interfaces. This PR is an alternative that answers that question, so the discussion has something concrete to compare against. Happy to close this in favour of #4732 if the maintainers prefer that route, or to fold this design into it.

## Design notes

**`QUERY` is not added to `IRoutes`.**
`IRoutes` is exported, so adding a method to it breaks every type outside of Gin that implements it: mocks, test doubles, wrappers, instrumentation layers. Those all fail to compile on upgrade, which makes a small feature a breaking release. Instead `QUERY` lives on the concrete `*RouterGroup` (and therefore on `*Engine`), and a separate interface is exported for the erased case:

```go
if q, ok := routes.(gin.IQueryRoutes); ok {
    q.QUERY("/search", searching)
}
```

`var _ IQueryRoutes = (*RouterGroup)(nil)` keeps that assertion honest at compile time. If the maintainers later decide a breaking change is acceptable, folding `IQueryRoutes` into `IRoutes` is a one-line change and the assertion above keeps compiling.

**`Any` is left alone.**
Adding `QUERY` to `anyMethods` silently registers an extra route for every existing `Any` call and changes what those handlers receive. Users who want it can be explicit:

```go
router.QUERY("/search", searching)
router.Match([]string{gin.MethodQuery, http.MethodPost}, "/search", searching)
```

**Why Gin declares the constant.**
`net/http` has no `http.MethodQuery` (golang/go#80058 is still open, verified against Go 1.25 and 1.26). `gin.MethodQuery` fills the gap and keeps the same value once the standard library catches up, so switching over later is mechanical and non-breaking.

**Binding works with no changes.**
`binding.Default` only special-cases `GET`, so a `QUERY` request with `Content-Type: application/json` resolves to the JSON binding and `c.ShouldBindJSON` reads the body as expected. Covered by a test.

## Tests

- `TestRouteQuery` — a `QUERY` route is invoked and binds its JSON body.
- `TestRouteQueryNotRegisteredByAny` — pins the documented `Any` behaviour.
- `TestQueryRoutesInterface` — registering through the `IQueryRoutes` assertion.
- `TestRouterGroupBasicHandle` — extended with `QUERY`, so groups and nested groups are covered like the other methods.
- `ginS.TestQUERY` — the package-level shortcut.

`go test ./...`, `go vet ./...` and `gofmt -l` are clean.


GP-09:
Note on the two red checks, neither is caused by this PR.

**Trivy Security Scan** — pre-existing on `master`. The scheduled scan has failed every night for at least the past week on the same commit, before this branch existed. Two HIGH findings in dependencies:

- `golang.org/x/text` v0.37.0, CVE-2026-56852, fixed in 0.39.0
- `golang.org/x/net` v0.55.0, CVE-2026-46600, fixed in 0.56.0

Opened #4807 to bump both, since it is a dependency change that deserves its own review rather than being buried in a feature PR. Trivy exits 0 with that branch. This PR goes green once it lands.

**codecov/project** — the report is comparing against a base that is 289 commits behind head, so the delta is noise from the stale baseline. `codecov/patch` is green: every line added here is covered by tests.

appleboy (approved):

PR #4817, opened 2026-08-27
fix(binding): optional size limit for BSON and Protobuf request bodies

### LINKED ISSUE #4759: BSON and Protobuf bindings read the entire request body without a size limit
## Summary

The BSON and Protobuf binders read the entire request body into memory before parsing it.

If an application uses these binders without an application-level or proxy-level body size limit, a large request body can force large server-side memory allocations and cause resource exhaustion.

## Affected Code

`binding/bson.go`

```go
func (b bsonBinding) Bind(req *http.Request, obj any) error {
    buf, err := io.ReadAll(req.Body)
    if err == nil {
        err = b.BindBody(buf, obj)
    }
    return err
}
```

`binding/protobuf.go`

```go
func (b protobufBinding) Bind(req *http.Request, obj any) error {
    buf, err := io.ReadAll(req.Body)
    if err != nil {
        return err
    }
    return b.BindBody(buf, obj)
}
```

Both paths call `io.ReadAll(req.Body)` without applying a maximum size.

## Reproduction

I reproduced the behavior against a minimal Gin server that directly calls the real `binding.BSON.Bind` path.

A small request body only caused a small allocation:

```text
binding.BSON.Bind err=invalid document length
HeapAlloc=0.7 MiB  TotalAlloc=0.7 MiB  (delta vs baseline = +0.1 MiB)
```

A 200 MB request body caused the request body to be fully read into memory before parsing failed:

```text
binding.BSON.Bind err=invalid document length
HeapAlloc=516.6 MiB  TotalAlloc=516.8 MiB  (delta vs baseline = +516.2 MiB)
```

The parse error happens after the full body has already been read, so the memory consumption is triggered even when the body is invalid.

The Protobuf binder uses the same `io.ReadAll(req.Body)` pattern.

## Expected Behavior

Binders should avoid unbounded request body reads, or Gin should provide a clear built-in way to enforce a maximum body size for these binders.

## Actual Behavior

The BSON and Protobuf binders read the full request body into memory without a size limit.

## Impact

A client can send a very large request body to an endpoint using these binders and force large memory allocations. This may lead to resource exhaustion or service degradation if no external request size limit is configured.

## Suggested Fix

- Add a configurable maximum body size for binders that need to buffer the full request body.
- Use a bounded reader before calling `io.ReadAll`.
- Document the default behavior clearly so applications using BSON or Protobuf binding know they must enforce request size limits.

### Gin Version

gin-gonic/gin@34dac20

Yes

### Source Code

```go
package main

import (
	"fmt"
	"net/http"
	"runtime"

	"github.com/gin-gonic/gin"
)

func main() {
	r := gin.Default()

	r.POST("/bson", func(c *gin.Context) {
		var before, after runtime.MemStats
		runtime.ReadMemStats(&before)

		var obj any
		err := binding.BSON.Bind(c.Request, &obj)

		runtime.ReadMemStats(&after)

		c.String(http.StatusOK,
			"binding.BSON.Bind err=%v\nHeapAlloc=%.1f MiB  TotalAlloc=%.1f MiB\n",
			err,
			float64(after.HeapAlloc)/1024/1024,
			float64(after.TotalAlloc)/1024/1024,
		)
	})

	if err := r.Run(":8080"); err != nil {
		panic(err)
	}
}
```

Run the server, then send a large request body:

```bash
go run .
head -c 209715200 /dev/zero | curl -X POST http://localhost:8080/bson --data-binary @-
```

Observed output from my reproduction:

```text
binding.BSON.Bind err=invalid document length
HeapAlloc=516.6 MiB  TotalAlloc=516.8 MiB
```

The BSON parse error occurs after the full request body has already been read into memory.

### Go Version

go1.26.5

### Operating System

Docker container based on golang:1.26

22elix3r:
## Summary

BSON and Protobuf binders call `io.ReadAll(req.Body)` with no size cap, so a large request can force a large allocation before parse fails. This adds `binding.MaxBodyBytes` so applications can bound those reads.

The default remains `0` (unlimited) to avoid a breaking change for existing large payloads. Issue discussion and earlier PRs disagreed on 32 MiB vs 128 MiB vs unlimited as a default.

## Related Issue (Fixes #4759)

Fixes #4759

## Changes Made

- Add `binding.MaxBodyBytes` and a shared `readBody` helper used by `binding.BSON.Bind` and `binding.ProtoBuf.Bind`.
- When `MaxBodyBytes > 0`, the body is read through `io.LimitReader`. Oversized bodies return `*http.MaxBytesError`, which `MustBindWith` already maps to HTTP 413.
- Document the knob in `docs/doc.md`.
- Tests call the shipped `BSON.Bind` and `ProtoBuf.Bind` for over-limit, within-limit, default-unlimited, and read-error paths.

## Testing

- `gofmt -w binding/body.go binding/bson.go binding/protobuf.go binding/binding_test.go`
- `go test -count=1 -timeout 180s ./binding`
- `go test -count=1 -timeout 180s .`
- `go vet ./binding .`
- `go test -count=1 -tags nomsgpack ./binding -run 'TestBindingBSONBodySizeLimit|TestBindingProtoBufBodySizeLimit|TestBindingBSON\$|TestBindingProtoBuf\$'`

```
ok  github.com/gin-gonic/gin/binding  0.007s
ok  github.com/gin-gonic/gin          0.179s
```

## Notes

Usage:

```go
binding.MaxBodyBytes = 32 << 20 // 32 MiB
```

JSON/XML/YAML/TOML binders already stream and are unchanged. Form binding already uses `ParseMultipartForm(defaultMemory)`.

This is intentionally opt-in. A hard 32 MiB default would reject legitimate large BSON/Protobuf bodies and is the point of disagreement on #4759 / #4785 / #4770 / #4767.


22elix3r:
Ready for review. codecov/project reported a drop; please treat it as informational unless the repo requires it. I cannot request a reviewer from this account.

22elix3r:
Withdrawing: I missed that this issue already had earlier open PRs. Sorry for the noise — leaving the field to those patches.

22elix3r:
Withdrawing this PR so the earlier open work on the issue can proceed.

PR #4820, opened 2026-08-28
fix(context): prevent nil pointer dereference in initFormCache when Request is nil (fix #4772)

### LINKED ISSUE #4772: PostForm getters panic when Context has no request

Jaymr1031-Z:
Fixes #4772

### Problem
When c.Request is nil (e.g., in isolated unit tests or uninitialized contexts), calling form getters like c.GetPostForm, c.PostForm, or c.DefaultPostForm triggers initFormCache(), which attempts to dereference 
eq.ParseMultipartForm(...) without a nil check, causing a panic: nil pointer dereference.

### Solution
- In context.go initFormCache(): Check if c.Request != nil before parsing multipart form data, matching the safety pattern in initQueryCache().
- Added test TestContextPostFormWithoutRequest in context_test.go.


context_test.go
```diff
@@ -3955,3 +3955,43 @@ func BenchmarkGetMapFromFormData(b *testing.B) {
...
+	assert.True(t, ok)
+	assert.Equal(t, "value3", val)
+
+	// Case 4: c.Request has multipart error, c.engine is nil
+	c4 := &Context{}
+	req4, _ := http.NewRequest(http.MethodPost, "/", strings.NewReader("bad multipart payload"))
+	req4.Header.Set("Content-Type", "multipart/form-data; boundary=boundary")
+	c4.Request = req4
+	_, ok = c4.GetPostForm("key")
+	assert.False(t, ok)
```
chenzeyan54-commits, line 3996: 💡 **Suggestion:** Consider adding `assert.Empty(t, c4.formCache)` to explicitly verify the cache state after a failed multipart parse — makes the intent clearer than relying solely on `assert.False(t, ok)`.

chenzeyan54-commits (approved):
## Code Review Summary

**Verdict: Approved** ✅

Clean defensive fix, well-tested, no regressions.

Two non-blocking observations:

1. **context_test.go:** Consider adding `assert.Empty(t, c4.formCache)` in Case 4 to explicitly verify the cache state after a failed multipart parse — makes the intent clearer than relying solely on `assert.False(t, ok)`.

2. **context.go (separate issue):** `FormFile()` (line 708) and `MultipartForm()` (line 724) have the same nil-deref pattern as the old `initFormCache`. Worth a separate issue/PR to add nil checks there too, for consistency.

PR #4825, opened 2026-09-02
fix(render): encode non-BMP characters as UTF-16 surrogate pairs in AsciiJSON

### LINKED ISSUE #4688: AsciiJSON silently corrupts non-BMP characters (emoji) by emitting malformed \u escapes
### Description

`render.AsciiJSON` corrupts any Unicode code point above U+FFFF (non-BMP characters such as emoji). It escapes every non-ASCII rune with `fmt.Appendf(buf, "\\u%04x", r)`, but `%04x` is a *minimum* width, not a fixed width. A non-BMP rune needs 5+ hex digits, so a single 5-digit `\uXXXXX` token is written. JSON `\u` escapes are **exactly 4 hex digits**, so a decoder reads the first 4 as one character and the trailing digit(s) as literal text.

The output is still *syntactically valid* JSON (so this is not a parse error and is easy to miss), but the value is silently wrong — it does not round-trip.

Per RFC 8259 §7, code points outside the BMP must be encoded as a UTF-16 surrogate pair (e.g. U+1F600 → `\ud83d\ude00`).

**Example — `😀` (U+1F600):**

| | value |
|---|---|
| Input | `😀` |
| `AsciiJSON` output | `{"msg":"\u1f600"}` |
| After `json.Unmarshal` | `ὠ0`  (U+1F60 `ὠ` + literal `0`) |
| Expected after round-trip | `😀` |

Offending code — `render/json.go` (`AsciiJSON.Render`):

```go
for _, r := range bytesconv.BytesToString(ret) {
    if r > unicode.MaxASCII {
        escapeBuf = fmt.Appendf(escapeBuf[:0], "\\u%04x", r) // 5+ digits for non-BMP → invalid escape
        buffer.Write(escapeBuf)
    } else {
        buffer.WriteByte(byte(r))
    }
}
```

Related (but distinct — that one is about naming/content-type/deprecation, not this corruption): #2546.

### Gin Version

`v1.12.0` (master, commit `d83f866`)

Yes.

Steps:
1. Render any payload containing a non-BMP character via `c.AsciiJSON(...)` (or `render.AsciiJSON`).
2. Observe the response body contains a malformed `\uXXXXX` escape.
3. `json.Unmarshal` the body — the decoded value differs from the original.

### Source Code

Minimal reproducible example:

```go
package main

import (
	"encoding/json"
	"fmt"
	"net/http/httptest"

	"github.com/gin-gonic/gin/render"
)

func main() {
	w := httptest.NewRecorder()
	_ = render.AsciiJSON{Data: map[string]string{"msg": "😀"}}.Render(w)

	out := w.Body.String()
	fmt.Println("rendered:", out) // {"msg":"\u1f600"}

	var decoded map[string]string
	_ = json.Unmarshal([]byte(out), &decoded)
	fmt.Printf("decoded: %q\n", decoded["msg"]) // "ὠ0", expected "😀"
}
```

### Go Version

`go1.25.0`

### Operating System

`Linux 6.17.0 x86_64` (codec-independent — the defect is in Gin's ASCII-escaping loop, not the JSON marshaler).

okxint:
## Problem

`AsciiJSON.Render` uses `fmt.Appendf(buf, "\\u%04x", r)` for every non-ASCII rune. The `%04x` verb is a **minimum** width — for a rune above `U+FFFF` (emoji, musical symbols, supplementary CJK…) it emits 5+ hex digits, producing an escape like `\u1f600`.

JSON `\u` escapes are **exactly 4 hex digits** (RFC 8259 §7). A decoder reads the first 4 digits as a different code point and leaves the remaining digit(s) as literal text. The output is still syntactically valid JSON, which is why this is easy to miss — values just silently fail to round-trip.

**Example — `😀` (U+1F600):**

| | value |
|---|---|
| Input | `😀` |
| `AsciiJSON` output (before) | `{"v":"\u1f600"}` |
| After `json.Unmarshal` | `ὠ0` — U+1F60 `ὠ` + literal `0` |
| Expected | `😀` |

Offending line (`render/json.go`):
```go
escapeBuf = fmt.Appendf(escapeBuf[:0], "\\u%04x", r)  // 5+ digits for non-BMP → invalid
```

## Fix

Detect runes `> U+FFFF` and split them into a UTF-16 surrogate pair per RFC 8259 §7:

```go
if r > 0xFFFF {
    r -= 0x10000
    high := 0xD800 + (r>>10)&0x3FF
    low := 0xDC00 + r&0x3FF
    escapeBuf = fmt.Appendf(escapeBuf[:0], "\\u%04x\\u%04x", high, low)
} else {
    escapeBuf = fmt.Appendf(escapeBuf[:0], "\\u%04x", r)
}
```

BMP characters (`U+0080`–`U+FFFF`) continue to use the single-escape path unchanged.

## Tests

Added `TestRenderAsciiJSONNonBMP` covering:

| Character | Code point | Expected pair |
|---|---|---|
| 😀 grinning face | U+1F600 | `\ud83d\ude00` |
| 𝄞 musical symbol G clef | U+1D11E | `\ud834\udd1e` |
| 𠀀 CJK Extension B (first) | U+20000 | `\ud840\udc00` |

Each case asserts:
1. The surrogate pair appears verbatim in the raw output.
2. `json.Unmarshal` recovers the original rune (full round-trip).

Closes #4688.


okxint:
Friendly ping. Happy to rebase or adjust anything if there's feedback.

PR #4826, opened 2026-09-02
fix: reset skippedNodes before each getValue call in HandleMethodNotAllowed loop

### LINKED ISSUE #4818: Panic: slice bounds out of range in getValue with HandleMethodNotAllowed enabled
### Description

With `HandleMethodNotAllowed = true`, gin can panic during route matching:                                                                                                                                                                                                           
```
runtime error: slice bounds out of range [:x+1] with capacity x
```                                            
                                                                                                                            
`c.skippedNodes` is allocated once per pooled Context with capacity `engine.maxSections` (`gin.go:254`). `getValue` pushes onto it with a raw reslice i.e. `(*skippedNodes)[:index+1]` (`tree.go:435`), so it cannot grow. `Context.reset()` clears it once per request (`context.go:117`), but `handleHTTPRequest` calls `getValue` once per method tree in the `HandleMethodNotAllowed` loop (`gin.go:746`) reusing the same stack, and `getValue` neither resets it on entry nor drains it before returning. Residue accumulates per tree until the reslice exceeds capacity:
```
getValue call #1: len=0 cap=6   push -> 1, push -> 2
getValue call #2: len=2 cap=6   <-- residue
getValue call #3: len=4 cap=6   <-- residue
getValue call #4: len=6 cap=6   push -> 7  => panic
```                                                                                                                                         

Because the panic happens during route matching, *before* the handler chain, `gin.Recovery()` cannot catch it, `net/http` recovers per connection, logs `http: panic serving ...`, and closes the connection with no response, so clients see an empty reply/reset rather than a 5xx.

### Gin Version

v1.12.0

Yes

### Source Code

gin v1.12.0, go 1.26.4

```go                                                                                                                                       
package main                                                                                                                                
                                                                                                                                            
import (                                                                                                                                    
      "net/http"                                                                                                                            
      "net/http/httptest"                                                                                                                   
      "testing"                                                                                                                             
                                                                                                                                            
      "github.com/gin-gonic/gin"                                                                                                            
)                                                                                                                                           
                                                                                                                                            
func TestPanic(t *testing.T) {                                                                                                              
      gin.SetMode(gin.ReleaseMode)                                                                                                          
      router := gin.New()                                                                                                                   
      router.HandleMethodNotAllowed = true                                                                                                  
                                                                                                                                            
      h := func(c *gin.Context) {}                                                                                                          
      router.OPTIONS("/:p0/:p1/a/:p2", h)                                                                                                   
      router.GET("/:p0/:p1/a/:p2", h)                                                                                                       
      router.PATCH("/b/:p0/:p1/c", h)                                                                                                       
      router.DELETE("/b/:p0/:p1/d/:p3", h)                                                                                                  
      router.GET("/b/:p0/:p1/e/f", h)                                                                                                       
      router.POST("/b/:p0/:p1/g/:p4/h", h)                                                                                                  
      router.OPTIONS("/b/:p0/:p1/g/:p4/h", h)                                                                                               
      router.DELETE("/b/cache", h)                                                                                                          
      router.GET("/b/clients/:p1/g", h)                                                                                                     
      router.POST("/b/clients/:p1/g", h)                                                                                                    
      router.PATCH("/b/clients/:p1/g/:p4", h)                                                                                               
      router.OPTIONS("/b/clients/:p1/g/:p4", h)
                                                                                                                                            
      req := httptest.NewRequest(http.MethodPost, "/b/clients/42", nil)                                                                     
      router.ServeHTTP(httptest.NewRecorder(), req)                                                                                         
}                                                                                                                                           
```

## Expectations                                                                                                                                                                                                                                                                           
`405 Method Not Allowed`.

## Actual result                                                                                                                  
```
panic: runtime error: slice bounds out of range [:7] with capacity 6                                                                        
github.com/gin-gonic/gin.(*node).getValue(...)                                                                                              
      gin@v1.12.0/tree.go:435                                                                                                               
github.com/gin-gonic/gin.(*Engine).handleHTTPRequest(...)                                                                                   
      gin@v1.12.0/gin.go:746                                                                                                                
github.com/gin-gonic/gin.(*Engine).ServeHTTP(...)                                                                                           
      gin@v1.12.0/gin.go:672                                                                                                                
```

### Go Version

go 1.26.4

### Operating System

_No response_

okxint:
## Problem

Closes #4818.

With `HandleMethodNotAllowed = true`, `handleHTTPRequest` iterates over every method tree to build the `Allow` response header. It passes the same `c.skippedNodes` slice to each `getValue` call:

```go
// gin.go:742-748 (before fix)
for _, tree := range engine.trees {
    if tree.method == httpMethod {
        continue
    }
    if value := tree.root.getValue(rPath, nil, c.skippedNodes, unescape); value.handlers != nil {
        allowed = append(allowed, tree.method)
    }
}
```

`getValue` extends the slice with a raw reslice (`tree.go:435`):

```go
*skippedNodes = (*skippedNodes)[:index+1]
```

The slice is allocated once per pooled `Context` with a fixed capacity of `engine.maxSections`. `Context.reset()` zeroes the length at request start, but **nothing resets it between `getValue` calls inside the loop**. With enough wildcard routes, residue from earlier trees fills the slice until a subsequent push exceeds capacity:

```
panic: runtime error: slice bounds out of range [:7] with capacity 6
    gin.(*node).getValue(...)      tree.go:435
    gin.(*Engine).handleHTTPRequest(...)  gin.go:746
    gin.(*Engine).ServeHTTP(...)   gin.go:672
```

Because the panic fires during route matching — before any handler executes — `gin.Recovery()` **cannot intercept it**. `net/http`'s per-connection recover logs `"http: panic serving …"` and tears down the connection with no response body; callers see an empty reply or a TCP reset instead of `405 Method Not Allowed`.

## Fix

Zero the slice length before each `getValue` call inside the `HandleMethodNotAllowed` loop:

```go
for _, tree := range engine.trees {
    if tree.method == httpMethod {
        continue
    }
    // Reset skippedNodes before each getValue call so residue from the
    // previous tree does not accumulate.
    *c.skippedNodes = (*c.skippedNodes)[:0]
    if value := tree.root.getValue(rPath, nil, c.skippedNodes, unescape); value.handlers != nil {
        allowed = append(allowed, tree.method)
    }
}
```

This mirrors exactly what `Context.reset()` already does at request start and costs one pointer write per method tree.

## Test

Added `TestMethodNotAllowedSkippedNodesPanic` — the exact route set and request from the issue report:

```go
router.HandleMethodNotAllowed = true
// 12 wildcard routes across OPTIONS/GET/PATCH/DELETE/POST
// ...
req := httptest.NewRequest(http.MethodPost, "/b/clients/42", nil)
assert.NotPanics(t, func() { router.ServeHTTP(w, req) })
assert.Equal(t, http.StatusMethodNotAllowed, w.Code)
```

Before the fix this test panics with `slice bounds out of range [:7] with capacity 6`. After the fix it passes and returns the correct `405`.


okxint:
Friendly ping. Happy to rebase or add tests if that would help move this forward.

PR #4827, opened 2026-09-05
fix(context): prevent nil pointer dereference in Cookie when Request is nil

leehainuo:
## Summary

`Context.Cookie` panics with a nil-pointer dereference when called on a `Context` whose `Request` is nil — the state produced by `CreateTestContext`, the standard way to unit-test middleware:

```go
c, _ := gin.CreateTestContext(httptest.NewRecorder()) // c.Request == nil
c.Cookie("foo") // panic: nil pointer dereference
```

Two existing conventions in this codebase already treat a request-less context as valid:

- the query getters were deliberately made nil-safe (`initQueryCache` guard + regression test `TestContextDefaultQueryOnEmptyRequest`);
- the binding layer rejects a nil request with an error instead of panicking (`binding/json.go`).

**Before:** panic. **After:** `("", http.ErrNoCookie)` — a request-less context has no cookies, so the natural result is the same sentinel the standard library returns when the named cookie is absent.

## Root cause

`Cookie(name)` dereferences `c.Request` directly (`c.Request.Cookie(name)`). This is the same bug class as #4772 (PostForm getters panic on nil request), but `Cookie` is **not** covered by the `initFormCache` guard proposed in the open #4772 PRs, so it needs its own guard.

## Changes

- `context.go`: nil-request guard in `Cookie(name)` returning `("", http.ErrNoCookie)`, with a comment explaining the semantics.
- `context_test.go`: `TestContextCookieWithNilRequest` placed right next to `TestContextDefaultQueryOnEmptyRequest` — asserts `Cookie` does not panic on a `CreateTestContext` context and returns `("", http.ErrNoCookie)`. It fails with a nil-pointer panic on the pre-fix code.

## Tests

- `make test` passes (exit 0); gin package coverage 98.7%.
- `go vet ./...` clean, `gofmt` clean.

# Pull Request Checklist

- [x] All tests pass in available continuous integration systems (e.g., GitHub Actions). *(locally: `make test` exit 0)*
- [ ] If the pull request introduces a new feature, the feature is documented in the `docs/doc.md`. *(N/A — bug fix, no new feature)*

Thank you for contributing!


wahajahmed010 (commented):
Nice catch on the nil-`Request` case. `CreateTestContext` is the documented way to unit-test middleware, so any helper that panics on the context it produces is going to bite someone. Returning `http.ErrNoCookie` is the right call — it matches what `(*http.Request).Cookie("missing")` already returns when the cookie is absent, so callers using `errors.Is(err, http.ErrNoCookie)` keep working without a new sentinel.

A few notes:

1. **Same gap probably exists on `SetCookieData`.** Just above the touched `Cookie`, `SetCookieData` does `if cookie == nil { return }` but then immediately calls `http.SetCookie(c.Writer, cookie)` (around line 1180ish depending on tree). If `c.Request` is nil there, `http.SetCookie` itself does not touch the request — it only reads the cookie — so it is probably fine. But `c.Cookie` is reached through several code paths and at least `c.Request.Cookie(...)` is the only one that touches the nil request. Worth a quick grep for other `c.Request.` reads in `context.go` to make sure this is the only place.

2. **Other nil-Request methods.** `FormValue`, `PostForm`, `DefaultQuery`, `Query`, `GetQuery`, `GetHeader`, `ClientIP`, `ContentType` all dereference `c.Request` directly. If you want this PR to stay minimal, that is fine, but a follow-up could either guard them or, better, document that `CreateTestContext` callers should set `c.Request = httptest.NewRequest(...)` before exercising those helpers. `TestContextDefaultQueryOnEmptyRequest` already exists next door and presumably tests one of these — would be worth checking whether it still passes with `c.Request == nil`.

3. **Test coverage.** `TestContextCookieWithNilRequest` covers exactly the panic case from the report, asserts the returned error wraps `http.ErrNoCookie`, and uses `assert.NotPanics` as a belt-and-braces guard. Good. One small thing: the test does not also assert behavior when `c.Request != nil` but the named cookie is missing — but that path is already covered elsewhere in this file, so the omission is fine.

4. **Minor: test comment placement.** The comment `// here c.Request == nil` is useful but is the only thing explaining why this test exists in isolation. Once a follow-up lands the broader nil-Request guard, this comment can move into a shared helper. Not blocking.

No changes required to land this; points 1 and 2 are the ones I would consider a follow-up issue rather than blockers on this PR.

PR #4828, opened 2026-09-06
fix(gin): reset skippedNodes on getValue entry to prevent 405 loop panic

### LINKED ISSUE #4818: Panic: slice bounds out of range in getValue with HandleMethodNotAllowed enabled

istoolsfox:
## Description

With `HandleMethodNotAllowed = true`, gin can panic during route matching (`runtime error: slice bounds out of range`), reproducible with the routes and request from #4818. The panic happens in the routing layer, before the handler chain, so `gin.Recovery()` cannot catch it — `net/http` recovers per connection, logs `http: panic serving ...`, and the client sees a closed connection with no response.

Fixes #4818

## Root cause

- `skippedNodes` is allocated once per pooled Context with a fixed capacity `engine.maxSections` (gin.go:254) and cleared once per request in `Context.reset()` (context.go:117).
- `getValue` pushes onto it with a raw reslice `(*skippedNodes)[:index+1]` (tree.go:435) — it cannot grow.
- On a **successful** match, `getValue` returns with pushed entries still in the slice (the pop loops only run on failure paths).
- The `HandleMethodNotAllowed` fallback loop (gin.go:746) calls `getValue` once per method tree reusing the same slice, so residue accumulates per tree until the reslice exceeds capacity — with the routes from the issue, `[:7] with capacity 6`.

## Fix

Start each `getValue` walk from an empty list:

```go
*skippedNodes = (*skippedNodes)[:0]
```

Skipped nodes are only meaningful within a single walk — they reference nodes of the tree being walked — and no caller reads the slice across calls (the other call site, the primary match on gin.go:715, is followed by either a handler or the same 405 loop). This makes `getValue` self-contained and defends all current and future call sites.

## Tests

`TestMethodNotAllowedSkippedNodesOverflow` registers the route table from #4818 and sends `POST /b/clients/42`: on master it panics with `slice bounds out of range [:7] with capacity 6`, with the fix it returns `404` (no other method tree matches this path) without panicking.


istoolsfox:
Note: the Trivy Security Scan failure on this branch is pre-existing on `master` (scheduled runs have been failing since 2026-08-07) — it flags CVE-2026-56854 in `golang.org/x/crypto` v0.53.0 (fixed in v0.55.0), a dependency that this diff does not touch. Happy to bump it in a separate commit/PR if that helps.

wahajahmed010 (commented):
Solid fix and good reproducer. The fact that the panic happens before `gin.Recovery()` and just closes the connection is exactly the class of bug a router should never have, so resetting `skippedNodes` at the top of `getValue` is the right place to fix it — the slice is genuinely per-walk state.

A few notes:

1. **Comment is more useful than the one-liner it replaced.** The block comment explaining the pooled-Context lifetime and the multi-method walk for `handleHTTPRequest` is exactly the kind of context that will save the next person from accidentally reintroducing the bug by trying to "optimize" the reset out. Good.

2. **Capacity is preserved across walks — is that intentional?** `*skippedNodes = (*skippedNodes)[:0]` keeps the underlying array and capacity. That is normally desirable (avoids re-allocation across the many small walks `handleHTTPRequest` issues). But if a previous walk on this `Context` ever pushed more entries than the current walk needs, the pool will hold onto the larger capacity indefinitely per Context. Given the `ContextPool` is reset between requests, this is fine in practice — just calling it out so reviewers know this is a deliberate choice and not an oversight.

3. **Reproducer test.** `TestMethodNotAllowedSkippedNodesOverflow` reproduces the exact route table from #4818, asserts both `NotPanics` and `StatusNotFound`, and references the issue. Good. One thing to consider adding: a follow-up assertion that calling the same `ServeHTTP` again on the same router does not regress (the slice-reset bug would in theory be order-dependent on the pooled slice state). The current test uses a fresh `httptest.NewRecorder()` and a fresh request, so it exercises one walk per Context — it would not catch a regression where the bug returns but only on the *second* request on a recycled Context. Not a blocker; mentioning it because the root cause is exactly Context-pool state leak.

4. **Coverage gap on the "second walk on same Context" case.** Tied to point 3: it would be cheap to add a sub-test that calls `router.ServeHTTP` twice in a row on the same router, with the second request being the one that triggers 405 logic, and asserts no panic. That would directly exercise the pool-reuse path. Optional.

5. **Other trees.** `tree.go` is shared by the regular and `httprouter`-style trees. Worth confirming there is no second `getValue` variant on a different type that also needs this reset; a quick `grep -n "skippedNodes" .` should be enough.

No blockers. The fix is minimal, the comment is good, and the regression test is well-targeted. Address point 5 if you have a moment.

istoolsfox:
Thanks for the thorough review! Point by point:

1. Comment — glad it reads well; kept as is.

2. Capacity retention is deliberate: the slice is allocated once per pooled `Context` with capacity `engine.maxSections` (gin.go:254) and cleared once per request in `Context.reset()` (context.go:117). Reusing the backing array across the many small walks `handleHTTPRequest` issues avoids re-allocation, and the per-request reset bounds any growth, so a recycled `Context` never carries a bloated slice into its next request. Happy to add a sentence to the comment if you'd like this spelled out in code.

3/4. Good catch — the single-walk test cannot see pool-reuse state. Added `TestMethodNotAllowedSkippedNodesOverflowOnContextReuse` (289b5a1): it drives the exact #4818 route table through two consecutive `router.ServeHTTP` calls on the same router, so the second 405 walk re-enters `getValue` on a recycled pooled `Context`, asserting no panic and 404 on both passes. Without the fix it fails with the same `slice bounds out of range [:7] with capacity 6` panic; with the reset it passes. The full suite shows no new failures (only the four pre-existing Windows file-permission failures also present on master).

5. Confirmed there is no second variant: `getValue` has a single definition (tree.go:418) with two call sites (gin.go:715 primary match, gin.go:746 405 fallback loop), and every `skippedNodes` reference is accounted for — allocation (gin.go:254), per-request reset (context.go:117), the entry reset introduced here (tree.go:425), the push (tree.go:440-442), and the unwind loops (tree.go:466, 597, 654). Nothing reads or carries the slice across walks outside these points.

appleboy:
fixed in https://github.com/gin-gonic/gin/pull/4819

PR #4748, opened 2026-07-21
fix(context): match Connection tokens for IsWebsocket

sonnemusk:
Use comma-separated Connection tokens instead of substring Contains("upgrade").

PR #4829, opened 2026-09-08
fix(context): recognize WebSocket upgrades in repeated headers

jakezwang:
`IsWebsocket` only checks the first `Connection` field and accepts any value containing `upgrade`. Inspect all field values and match complete, case-insensitive tokens so repeated headers work and values such as `notupgrade` are rejected.

The regression table covers repeated fields, token boundaries, whitespace, casing, and existing valid handshakes. Five cases fail before the fix. `go test -race ./...` and the repository's golangci-lint 2.11 checks pass.

Related: #4748, which was withdrawn by its author. This change also covers repeated header fields.

Rebased onto current master, including its dependency security update. On Go 1.26.4, full race tests, the `nomsgpack`, `go_json`, and `sonic` variants, `go vet`, and golangci-lint 2.11.4 pass. Updated-branch CI passes all test variants, lint, CodeQL, and Trivy. Codecov covers all modified lines; its project check still compares against a baseline 296 commits behind master (98.64% versus 99.21%).

# Pull Request Checklist

- [ ] All tests pass in available continuous integration systems (e.g., GitHub Actions).


wahajahmed010:
Nice fix. The previous `strings.Contains(..., "upgrade")` was a textbook tokenization bug — it would have accepted `notupgrade` and any string with the substring "upgrade". The rewrite using `strings.SplitSeq` and `strings.EqualFold` is the correct shape.

A few notes:

1. **Token trimming is only ` 	`** — RFC 7230 says optional whitespace can surround tokens in a comma-separated list, including HTAB and SP. Your trim covers both, which is correct. The non-HTTP-whitespace case ` ` (non-breaking space) is correctly rejected by your test, since ` upgrade` is not a valid HTTP token per spec.

2. **`strings.SplitSeq` is Go 1.24+** — I see from the import block (Go version) that gin supports recent Go versions. If gin still claims Go 1.21 or 1.22 compatibility, `SplitSeq` will fail to build. Worth confirming `go.mod` minimum is at least 1.24.

3. **Order of checks inverted from the original** — you check `Upgrade: websocket` first, then iterate `Connection`. This is fine because both must be true for a websocket handshake, but it changes the short-circuit behavior: previously a missing `Upgrade: websocket` would still iterate `Connection` headers (cheap, but more work). The new version short-circuits on `Upgrade` first, which is more efficient. Just noting for future maintainers reading the diff.

4. **Test coverage** — the regression table is solid: token boundaries (prefix, suffix), whitespace, case-insensitivity, repeated fields, empty first field, non-HTTP whitespace, and the existing valid handshake. Good. One edge case worth adding: a `Connection` field with only commas, e.g. `,,`, which would split into empty tokens. Your trim would produce empty strings that fail `EqualFold(..., "upgrade")`, so the test would still pass — but a dedicated test makes that contract explicit.

5. **Codecov base branch issue** — your note about the base being 292 commits behind master for the project coverage check is concerning. If the project-level threshold is enforced on a stale base, future contributors will see the same false-positive failure. Worth filing a tracking issue to update the Codecov base target.

Approving.

